PR #11, opened 2020-03-29
Added the potentiometer-element.ts component

gilf:
Added a new potentiometer-element.ts component and it's story to the component library.
Let me know what do you think.

### DIFF
--- a/src/potentiometer-element.stories.ts
+++ b/src/potentiometer-element.stories.ts
@@ -10,7 +10,7 @@ storiesOf('potentiometer', module).add(
       <wokwi-potentiometer
         min="0"
         max="200"
-        @potentiometer-value-changed=${action('potentiometer-value-changed')}
+        @input=${action('potentiometer-value-changed')}
       ></wokwi-potentiometer>
     `
 );
--- a/src/potentiometer-element.stories.ts
+++ b/src/potentiometer-element.stories.ts
@@ -8,8 +8,8 @@ storiesOf('potentiometer', module).add(
   () =>
     html`
       <wokwi-potentiometer
-        @min="0"
-        @max="200"
+        min="0"
+        max="200"
         @potentiometer-value-changed=${action('potentiometer-value-changed')}
       ></wokwi-potentiometer>
     `
--- a/src/potentiometer-element.stories.ts
+++ b/src/potentiometer-element.stories.ts
@@ -1,0 +1,16 @@
+import { storiesOf } from '@storybook/web-components';
+import { action } from '@storybook/addon-actions';
+import { html } from 'lit-html';
+import './potentiometer-element';
+
+storiesOf('potentiometer', module).add(
+  'potentiometer',
+  () =>
+    html`
+      <wokwi-potentiometer
+        @min="0"
+        @max="200"
+        @potentiometer-value-changed=${action('potentiometer-value-changed')}
+      ></wokwi-potentiometer>
+    `
+);
--- a/src/potentiometer-element.ts
+++ b/src/potentiometer-element.ts
@@ -175,6 +175,6 @@ export class PotentiometerElement extends LitElement {
     const updated = Math.round(clamped / this.step) * this.step;
     const rounded = Math.round(updated * 100) / 100;
     this.updatePotentiometerPointer(rounded);
-    this.dispatchEvent(new InputEvent('potentiometer-value-changed', { data: `${rounded}` }));
+    this.dispatchEvent(new InputEvent('input', { data: `${rounded}` }));
   }
 }
--- a/src/potentiometer-element.ts
+++ b/src/potentiometer-element.ts
@@ -1,0 +1,269 @@
+import { customElement, html, LitElement, property } from 'lit-element';
+
+interface Point {
+  x: number;
+  y: number;
+}
+
+let id = 1;
+
+/** The potentiometer SVG is taken from https://freesvg.org/potentiometer and some of the
+    functions are taken from https://github.com/vitaliy-bobrov/js-rocks knob component */
+@customElement('wokwi-potentiometer')
+export class PotentiometerElement extends LitElement {
+  @property() min = 0;
+  @property() max = 100;
+  @property() value = 0;
+  @property() step = 1;
+  @property() startDegree = -135;
+  @property() endDegree = 135;
+  private center: Point = { x: 0, y: 0 };
+  pressed = false;
+  id = `rotating-${id++}`;
+
+  clamp(min: number, max: number, value: number): number {
+    return Math.min(Math.max(value, min), max);
+  }
+
+  mapToMinMax(value: number, min: number, max: number): number {
+    return value * (max - min) + min;
+  }
+
+  percentFromMinMax(value: number, min: number, max: number): number {
+    return (value - min) / (max - min);
+  }
+
+  render() {
+    return html`
+      <style>
+        #${this.id} {
+          transform-origin: -120.168198px -140.5804px;
+          transform: rotate(var(--knob-angle, 0deg));
+        }
+      </style>
+      <svg
+        xmlns="http://www.w3.org/2000/svg"
+        width="20mm"
+        height="20mm"
+        viewBox="0 0 20 20"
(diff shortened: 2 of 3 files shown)


src/potentiometer-element.ts
```diff
@@ -0,0 +1,269 @@
...
+        height="20mm"
+        viewBox="0 0 20 20"
+        @mousedown=${this.down}
+        @mousemove=${this.move}
+        @mouseup=${this.up}
+        @touchstart=${this.down}
+        @touchmove=${this.move}
+        @touchend=${this.up}
+      >
+        <g transform="translate(-6.7420057,-6.6693388)">
```
urish: Can you remove these transforms?

This is how I do it: open the SVG file in inkscape, select the group (you can simply click on one of the elements, or press Ctrl+Shift+X to open the XML explorer, and then select it from the XML explorer), and then ungroup (Ctrl+U). 

Ungrouping removes these transforms, and changes the x/y (and possibly heigh/width) of the child elements, so that they already have the transforms applied.

src/potentiometer-element.ts
```diff
@@ -0,0 +1,269 @@
...
+          <g transform="matrix(0.15090139,0,0,0.15090139,34.862601,35.84314)">
+            <rect
+              ry="8.1522207"
+              y="-192.35448"
+              x="-185.37505"
+              height="128.89323"
+              width="129.11356"
+              style="opacity:1;fill:#045881;fill-opacity:1;stroke:#045881;stroke-width:1.95151615;stroke-linecap:butt;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
```
urish: Can you remove all the extraneous styles?  Stuff like `opacity:1` and `fill-opacity:1` is already the default. Basically, I think the only ones that you need to keep (without affecting the results) are:
- fill
- stroke
- stroke-width (only if stroke is not set to `none`). You can also round it (1.95 is enough, no need for all those extra decimal digits)
- opacity (only if < 1)
- stroke-opacity (only if < 1, stroke is not set to `none`)

This will clean up a lot of the clutter.

src/potentiometer-element.ts
```diff
@@ -0,0 +1,269 @@
...
+              width="42.74408"
+              style="opacity:1;fill:#000000;fill-opacity:0;stroke:#ffffff;stroke-width:1.95151615;stroke-linecap:butt;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
+            />
+            <text
+              y="-83.291"
+              x="-145.16521"
+              style="font-style:normal;font-weight:normal;font-size:6.4255476px;line-height:1.25;font-family:sans-serif;letter-spacing:0px;word-spacing:0px;fill:#ffffff;fill-opacity:1;stroke:none;stroke-width:0.34422573"
+              xml:space="preserve"
+            >
+              <tspan
```
urish: The `<tspan>`s are redundant (they are only needed if you want to have several spans of texts with different styles / positions). You can remove them, and just copy their `x`, `y`, and styles to the enclosing `<text>` element (so basically <text bla bla>GND</text>.

src/potentiometer-element.ts
urish: Also, make sure to do this before doing any of the other SVG changes I suggested (removing styles, tspans, etc.), because inkscape is likely to add them again

src/potentiometer-element.ts
```diff
@@ -0,0 +1,269 @@
...
+                x="-174.16521"
+              >
+                GND
+              </tspan>
+            </text>
+            <text
+              y="-83.155312"
+              x="-128.39781"
+              style="font-style:normal;font-weight:normal;font-size:6.42554712px;line-height:1.25;font-family:sans-serif;letter-spacing:0px;word-spacing:0px;fill:#ffffff;fill-opacity:1;stroke:none;stroke-width:0.34422573"
+              xml:space="preserve"
```
urish: Similar to the graphics items, most of the styles here are redundant, as the `font-style` and `font-weight` are already the default.

I'd move all the font related styles (font-size, line-height, font-family, etc.) to the top SVG element, so that they won't have to be repeated for each `text` element. 

And the font size can be shortened to just `6.4`, without all the trailing decimals... (Inkscape just loves them!)

src/potentiometer-element.ts
```diff
@@ -0,0 +1,269 @@
...
+      x: window.scrollX + potentiometerRect.left + potentiometerRect.width / 2,
+      y: window.scrollY + potentiometerRect.top + potentiometerRect.height / 2
+    };
+  }
+
+  private rotateHandler(event: MouseEvent | TouchEvent) {
+    event.stopPropagation();
+    event.preventDefault();
+
+    const isTouch = event.type === 'touchmove';
```
urish: Tested on mobile, works like a charm! 😍

src/potentiometer-element.ts
```diff
@@ -0,0 +1,269 @@
...
+    const deg = Math.round((this.endDegree - this.startDegree) * percent + this.startDegree);
+    this.style.setProperty('--knob-angle', `${deg}deg`);
+  }
+
+  private updateValue(value: number) {
+    const clamped = this.clamp(this.min, this.max, value);
+    const updated = Math.round(clamped / this.step) * this.step;
+    const rounded = Math.round(updated * 100) / 100;
+    this.updatePotentiometerPointer(rounded);
+    this.dispatchEvent(new CustomEvent('potentiometer-value-changed', { detail: rounded }));
```
urish: How about dispatching an [`InputEvent`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event)? 
This will make the behavior of the element more in line with a standard  `<input>` element

src/potentiometer-element.stories.ts
```diff
@@ -0,0 +1,16 @@
...
+import { html } from 'lit-html';
+import './potentiometer-element';
+
+storiesOf('potentiometer', module).add(
+  'potentiometer',
+  () =>
+    html`
+      <wokwi-potentiometer
+        @min="0"
+        @max="200"
```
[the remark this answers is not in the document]
gilf: Yep. My bad :)

src/potentiometer-element.ts
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  percentFromMinMax(value: number, min: number, max: number): number {
+    return (value - min) / (max - min);
+  }
+
+  render() {
+    return html`
+      <style>
+        #${this.id} {
```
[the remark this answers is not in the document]
gilf: No problem. Will be chnaged

src/potentiometer-element.ts
```diff
@@ -0,0 +1,269 @@
...
+    const percent = this.percentFromMinMax(deg, this.startDegree, this.endDegree);
+    const value = this.mapToMinMax(percent, this.min, this.max);
+
+    this.updateValue(value);
+  }
+
+  private updatePotentiometerPointer(value: number) {
+    const percent = this.percentFromMinMax(value, this.min, this.max);
+    const deg = Math.round((this.endDegree - this.startDegree) * percent + this.startDegree);
+    this.style.setProperty('--knob-angle', `${deg}deg`);
```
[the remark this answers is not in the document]
gilf: I'm used to working with CSS variables so I used this approach.

src/potentiometer-element.ts
gilf: On it

src/potentiometer-element.ts
gilf: Removed

src/potentiometer-element.ts
gilf: Done

src/potentiometer-element.ts
gilf: :)

src/potentiometer-element.ts
gilf: InputEvent it will be

src/potentiometer-element.ts
gilf: Done

PR #13, opened 2020-04-03
feat(ssd1306): added 128X64 OLED

LironHazan:


src/ssd1306-element-element.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    const imageData = new ImageData(arr, 200);
+
+    ctx.putImageData(imageData, 0, 0);
+  }
+
+  firstUpdated() {
+    /// this.putGivenImageData();
+    // Getting null for canvas, there's an issue here
+    this.testImageData();
```
LironHazan, line 40: @urish maybe you'll have an idea, at this point I expected to have the rendered DOM with the canvas element, but querying it will resolve as null so I can't test the image data, (+ the size of the foreignObject and hosted canvas isn't right as well)

https://codepen.io/lironhazan/pen/VwLJgYK

src/ssd1306-element-element.ts
```diff
@@ -1,12 +1,44 @@
...
 
 @customElement('wokwi-ssd1306-element')
 export class Ssd1306ElementElement extends LitElement {
-  @property() dimensions: NumTuple = [141, 108];
-  @property() bitmap: any; // size should be 128 x 64
+  @property() dimensions: Dimensions = [141, 108];
+  @property() imageData: ImageData = new ImageData(128, 64);
+
+  putGivenImageData() {
+    const canvas: HTMLCanvasElement | null = document.querySelector('canvas');
```
urish, line 12: With `lit-element`, the components are enclosed in Shadow DOM, so there is an API for accessing them, called `shadowRoot`. You can see an example here: https://lit-element.polymer-project.org/guide/lifecycle#examples

src/ssd1306-element-element.ts
```diff
@@ -1,12 +1,44 @@
...
 import { customElement, html, LitElement, property, SVGTemplateResult } from 'lit-element';
 
-type NumTuple = [number, number];
+type Dimensions = [number, number];
 
 @customElement('wokwi-ssd1306-element')
 export class Ssd1306ElementElement extends LitElement {
-  @property() dimensions: NumTuple = [141, 108];
-  @property() bitmap: any; // size should be 128 x 64
+  @property() dimensions: Dimensions = [141, 108];
```
urish, line 8: It'd be better to have separate `width` / `height` properties, this way they can also be set via plain HTML when creating an instance of you component (also, it's pretty standard in HTML, e.g. the ``, `<svg>` and `<canvas>` tags all follow this convention)

src/ssd1306-element-element.ts
```diff
@@ -24,6 +56,10 @@ export class Ssd1306ElementElement extends LitElement {
         <!-- 128 x 64 screen -->
         <path fill="#1A1A1A" d="M21 26h100v54H21z" />
 
+        <foreignObject transform="translate(4 6)" x="20" y="20" width="128" height="64">
```
urish, line 59: Is there a reason we need both `translate` and `x` / `y`?

src/ssd1306-element-element.ts
LironHazan: done

src/ssd1306-element-element.ts
LironHazan: nope I fixed :)

src/ssd1306-element-element.ts
LironHazan: thanks :) it works

LironHazan:
@urish I had some proportions issues with the original size (not 100% sure why it happened, I used the original metrics..) so I stayed loyal to the oled screen size and did some small changed to the svg, LMK what do you think? I changed the width/hight accessors to be fixed to a const size instead of being a user input (cause I don't draw the svg dynamically at this point, an interesting challenge though...)

src/ssd1306-128x64-element.ts
```diff
@@ -0,0 +1,93 @@
...
+          </foreignObject>
+        </g>
+
+        <!-- All texts -->
+        <g
+          fill="#FFF"
+          text-anchor="middle"
+          font-size="5"
+          font-weight="300"
+          font-family="MarkerFelt-Wide, Marker Felt"
```
urish, line 49: These fonts are not available on all computers - let's add a fallback from one of the web-safe fonts (https://www.w3schools.com/cssref/css_websafe_fonts.asp)

LironHazan: done

src/ssd1306-128x64-element.ts
```diff
@@ -0,0 +1,100 @@
...
+  private screenHeight = 64;
+  @property() imageData: ImageData;
+
+  constructor() {
+    super();
+    this.imageData = new ImageData(this.screenWidth, this.screenHeight);
+  }
+
+  putGivenImageData() {
+    const canvas: HTMLCanvasElement | null | undefined = this.shadowRoot?.querySelector('canvas');
```
urish, line 18: Could possibly be written as
```
    const canvas = this.shadowRoot?.querySelector<HTMLCanvasElement>('canvas');
```

src/ssd1306-element.ts
```diff
@@ -0,0 +1,114 @@
...
+          </g>
+
+          <g transform="translate(41 17)">
+            <text x="0" y="6">C1k</text>
+            <text x="12" y="6">DC</text>
+            <text x="23" y="6">Rst</text>
+            <text x="39" y="6">3v3</text>
+            <text x="58" y="6">Gnd</text>
+          </g>
+          <!--Start -->
```
urish, line 93: Star?

LironHazan: yes fixed

src/ssd1306-element.stories.ts
```diff
@@ -0,0 +1,18 @@
...
+storiesOf('Ssd1306 Element', module)
+  .addParameters({ component: 'wokwi-ssd1306-element' })
+  .addDecorator(withKnobs)
+  .add('Default', () => html` <wokwi-ssd1306-element></wokwi-ssd1306-element> `)
+  .add(
+    'Request new image data',
+    () =>
+      html`
+        <wokwi-ssd1306-element
+          updateImage="${boolean('updateImage', false)}"
```
urish, line 15: We no longer have `updateImage` :-)

src/ssd1306-element.stories.ts
```diff
@@ -0,0 +1,18 @@
+import { withKnobs, boolean } from '@storybook/addon-knobs';
+import { storiesOf } from '@storybook/web-components';
+import { html } from 'lit-html';
+import './ssd1306-element';
+
+storiesOf('Ssd1306 Element', module)
```
urish, line 6: We can remove "Element" from the name (just to stay consistent with the other stories), also call it SSD1306 (like we used uppercase in LED and LCD1602)

PR #16, opened 2020-04-11
fix(potentiometer): added keyboard reactions

LironHazan:


urish:
Works like a charm! Two things are missing though:

1. We should visually indicate when the element has a focus. The standard way to do it is using the [`outline` CSS property](http://www.outlinenone.com/)
2. Clicking on the element should focus it (currently the only way to focus it is by using the tab key)

LironHazan:
> Works like a charm! Two things are missing though:
> 
> 1. We should visually indicate when the element has a focus. The standard way to do it is using the [`outline` CSS property](http://www.outlinenone.com/)
> 2. Clicking on the element should focus it (currently the only way to focus it is by using the tab key)

@urish done, I used stroke instead of the outline for the rotating part of the knob (rect)

src/potentiometer-element.ts
```diff
@@ -120,6 +147,26 @@ export class PotentiometerElement extends LitElement {
     `;
   }
 
+  observeInputInteractions() {
+    this.inputEl = this.shadowRoot?.querySelector('.hide-input');
+    const potentiometerKnob: SVGRectElement | null | undefined = this.shadowRoot?.querySelector(
+      '#rotating'
+    );
+
+    this.inputEl?.addEventListener('focus', () => {
```
urish, line 156: You can probably achieve the same with css, e.g. something like:

`input:focus + svg #rotating` or similar

LironHazan: Thanks I actually tried input:focus + svg #rotating in one of my tries and it didn't work but maybe I selected it by class instead of id.. I'll check it out again :) 
regarding the lit-element handlers totally right I wasn't thinking :)

src/potentiometer-element.ts
```diff
@@ -81,6 +112,7 @@ export class PotentiometerElement extends LitElement {
         />
         <rect x="5.4" y=".70" width="9.1" height="1.9" fill="#ccdae3" stroke-width=".15" />
         <ellipse
+          filter="url(#outline)"
```
urish, line 115: How about controlling this filter directly from CSS? then you don't really need the input event listeners and updating the stdDeviation from JS

LironHazan: always better - will fix :)

urish:
Released as version 0.13.1

PR #19, opened 2020-04-12
improvement: style of pressed buttons

karinchechik:


src/membrane-keypad-element.ts
```diff
@@ -26,27 +33,46 @@ export class MembraneKeypadElement extends LitElement {
...
-        g[tabindex]:focus {
-          fill: #4e50d7;
+        g[tabindex]:focus,
+        g[tabindex]:active {
           stroke: white;
           outline: none;
         }
 
+        .blueKey:focus,
+        .redKey:focus {
```
urish, line 57: We use "kebab-case" for CSS class name in [other parts of the project](https://github.com/wokwi/wokwi-elements/blob/1467d953010919da5d51fc2bfa692a04ae7e8983/src/lcd1602-element.ts#L29), so let's change these to `blue-key` and `red-key` to be consistent

src/membrane-keypad-element.ts
```diff
@@ -129,8 +160,9 @@ export class MembraneKeypadElement extends LitElement {
     }
   }
 
-  private up(key: string) {
+  private up(key: string, event: any) {
```
urish, line 163: You can either use `UIEvent` or `KeyboardEvent | MouseEvent | TouchEvent` (I personally prefer the first one)

src/membrane-keypad-element.ts
```diff
@@ -8,16 +8,23 @@ export class MembraneKeypadElement extends LitElement {
 
   private pressedKeys = new Set<string>();
 
+  isNumeric(text: string) {
+    return !isNaN(parseFloat(text));
+  }
+
   renderKey(text: string, x: number, y: number) {
+    const keyStyle = this.isNumeric(text) ? 'blueKey' : 'redKey';
```
urish, line 16: Maybe a better name would be `keyClass` or `keyClasses`?

src/membrane-keypad-element.ts
```diff
@@ -8,16 +8,23 @@ export class MembraneKeypadElement extends LitElement {
 
   private pressedKeys = new Set<string>();
 
+  isNumeric(text: string) {
```
urish, line 11: should be private (or you can just declare it as a plain local function at the start of the file, so it's not even a part of the class)

src/membrane-keypad-element.ts
karinchechik: To use either option, I had to declare 'event.currentTarget as SVGElement' to avoid TS error: Property 'classList' does not exist on type 'EventTarget'. I updated with the first option - UIEvent.

src/membrane-keypad-element.ts
```diff
@@ -118,8 +148,10 @@ export class MembraneKeypadElement extends LitElement {
     `;
   }
 
-  private down(key: string) {
+  private down(key: string, event: UIEvent) {
     if (!this.pressedKeys.has(key)) {
+      const currTarget = event.currentTarget as SVGElement;
+      currTarget.classList.add('pressed');
```
urish, line 154: What's the reason you need the `pressed` class?
It seems like the `:active` selector that you have above already does the trick

karinchechik: The active selector isn't triggered when pressing the space key, so I added this class. 
But it has an isssue - while pressing space, if you press tab and press space again or click with the mouse on another button, it will display multiple buttons as pressed. I can fix the tab option by removing the class when pressing tab. But I couldn't think of a way to fix the mouse selection. Do you think I should check every button on mousedown? Got a better idea?

urish: Good point! 

So perhaps you could:
1. For the tab issue - remove the `pressed` class on `blur` event. That's way, when you press the tab key, the element loses focus, and the highlight is gone.
2. Make the event parameter optional and only pass it on keyboard event, so it won't get added on mouse events (this will probably solve the issue with the class sticking around when you click on the button, and then drag the mouse pointer out of the element area, which the button is still pressed)

PR #22, opened 2020-04-16
improvement: added keyboard support

karinchechik:
I added the option to press the numbers, letters and the signs of the membarne keypad on the keyboard.

### DIFF
--- a/src/membrane-keypad-element.ts
+++ b/src/membrane-keypad-element.ts
@@ -1,6 +1,15 @@
 import { customElement, html, LitElement, property, svg } from 'lit-element';
 
 const SPACE_KEY = 32;
+const SHIFT_KEY = 16;
+
+const customKeydownEvent = new KeyboardEvent('keydown', {
+  code: 'Space',
+});
+
+const customKeyupEvent = new KeyboardEvent('keyup', {
+  code: 'Space',
+});
 
 function isNumeric(text: string) {
   return !isNaN(parseFloat(text));
@@ -11,14 +20,22 @@ export class MembraneKeypadElement extends LitElement {
   @property() threeColumns = false;
 
   private pressedKeys = new Set<string>();
+  private shiftPressed = false;
 
   renderKey(text: string, x: number, y: number) {
     const keyClass = isNumeric(text) ? 'blue-key' : 'red-key';
+    let keyId = `key${text}`;
+    if (text === '*') {
+      keyId = 'keyAsterisk';
+    } else if (text === '#') {
+      keyId = 'keyHash';
+    }
 
     return svg`<g
       transform="translate(${x} ${y})"
       tabindex="0"
       class=${keyClass}
+      id=${keyId}
       @blur=${(e: FocusEvent) => {
         this.up(text, e);
       }}
@@ -26,8 +43,10 @@ export class MembraneKeypadElement extends LitElement {
       @mouseup=${() => this.up(text)}
       @touchstart=${() => this.down(text)}
       @touchend=${() => this.up(text)}
-      @keydown=${(e: KeyboardEvent) => e.keyCode === SPACE_KEY && this.down(text, e)}
-      @keyup=${(e: KeyboardEvent) => e.keyCode === SPACE_KEY && this.up(text, e)}
+      @keydown=${(e: KeyboardEvent) =>
+        (e.keyCode === SPACE_KEY || e.code === 'Space') && this.down(text, e)}
+      @keyup=${(e: KeyboardEvent) =>
+        (e.keyCode === SPACE_KEY || e.code === 'Space') && this.up(text, e)}
     >
       <use xlink:href="#key" />
       <text x="5.6" y="8.1">${text}</text>
@@ -92,6 +111,8 @@ export class MembraneKeypadElement extends LitElement {
         font-size="8.2px"
         text-anchor="middle"
         xmlns="http://www.w3.org/2000/svg"
+        @keydown=${(e: KeyboardEvent) => this.keyStrokeDown(e.keyCode)}
+        @keyup=${(e: KeyboardEvent) => this.keyStrokeUp(e.keyCode)}
       >
         <defs>
           <rect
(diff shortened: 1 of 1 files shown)


src/membrane-keypad-element.ts
```diff
@@ -182,4 +203,81 @@ export class MembraneKeypadElement extends LitElement {
...
+    return keyName;
+  }
+
+  private keyStrokeDown(key: number) {
+    let selectedKey;
+    const keyName = this.defineKey(key);
+
+    if (keyName) {
+      selectedKey = this.shadowRoot?.querySelector(keyName);
+      selectedKey?.dispatchEvent(customKeydownEvent);
```
urish, line 264: How about just calling `down()` and `up()` directly to simplify things? We can change their signature to access the target element instead of an event object (after all, they only need the target element, not the entire event).

src/membrane-keypad-element.ts
```diff
@@ -11,23 +20,33 @@ export class MembraneKeypadElement extends LitElement {
...
+  private shiftPressed = false;
 
   renderKey(text: string, x: number, y: number) {
     const keyClass = isNumeric(text) ? 'blue-key' : 'red-key';
+    let keyId = `key${text}`;
+    if (text === '*') {
+      keyId = 'keyAsterisk';
+    } else if (text === '#') {
+      keyId = 'keyHash';
+    }
```
urish, line 32: I wonder if there is a way to do this that won't break if we add a new, non-numeric key in the future. Low priority, but if you have an idea that doesn't require specific handling for Asterisk and Hash, that would be nice.

src/membrane-keypad-element.ts
```diff
@@ -11,23 +20,33 @@ export class MembraneKeypadElement extends LitElement {
   @property() threeColumns = false;
 
   private pressedKeys = new Set<string>();
+  private shiftPressed = false;
 
   renderKey(text: string, x: number, y: number) {
     const keyClass = isNumeric(text) ? 'blue-key' : 'red-key';
+    let keyId = `key${text}`;
```
urish, line 27: Let's keep it consistent with the other elements on use kebab-case instead of camelCase (i.e. key-A instead of keyA)

src/membrane-keypad-element.ts
```diff
@@ -92,6 +111,8 @@ export class MembraneKeypadElement extends LitElement {
         font-size="8.2px"
         text-anchor="middle"
         xmlns="http://www.w3.org/2000/svg"
+        @keydown=${(e: KeyboardEvent) => this.keyStrokeDown(e.keyCode)}
```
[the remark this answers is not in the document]
karinchechik: Thanks, that works much better!

src/membrane-keypad-element.ts
```diff
@@ -11,23 +20,33 @@ export class MembraneKeypadElement extends LitElement {
...
         this.up(text, e);
       }}
       @mousedown=${() => this.down(text)}
       @mouseup=${() => this.up(text)}
       @touchstart=${() => this.down(text)}
       @touchend=${() => this.up(text)}
-      @keydown=${(e: KeyboardEvent) => e.keyCode === SPACE_KEY && this.down(text, e)}
-      @keyup=${(e: KeyboardEvent) => e.keyCode === SPACE_KEY && this.up(text, e)}
+      @keydown=${(e: KeyboardEvent) =>
+        (e.keyCode === SPACE_KEY || e.code === 'Space') && this.down(text, e)}
```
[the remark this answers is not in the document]
karinchechik: I used a cutom keyboardEvent, and since keyCode is deprecated it can't be defined, so I used code instead. But now I removed the custom event, and I'm not using e.code anymore.

src/membrane-keypad-element.ts
karinchechik: I removed this part. It is problematic to use '#' in an id value, so now I used data attribute instead.

src/membrane-keypad-element.ts
```diff
@@ -182,4 +187,26 @@ export class MembraneKeypadElement extends LitElement {
...
+    const selectedKey = this.shadowRoot?.querySelector(`[data-key-name="${keyName}"]`);
+    this.down(text, selectedKey as SVGElement);
+  }
+
+  private keyStrokeUp(key: string) {
+    let selectedKey = this.shadowRoot?.querySelector('.pressed');
+    const keyName = `key-${key.toLowerCase()}`;
+
+    if (key === 'Shift') {
+      this.shiftPressed = false;
```
urish, line 203: seems like `shiftPressed` is only set to false here, and never read. Perhaps we can remove it altogether?

src/membrane-keypad-element.ts
```diff
@@ -182,4 +187,26 @@ export class MembraneKeypadElement extends LitElement {
...
+      selectedKey?.classList.remove('pressed');
```
urish, line 204: Actually, this means that if I click on `1`, then press `Shift` and leave it, it will seem like `1` is no longer pressed, even though the actual release event will only be fired when I release the `1` key (which is the expected behavior)

src/membrane-keypad-element.ts
```diff
@@ -182,4 +187,26 @@ export class MembraneKeypadElement extends LitElement {
...
+      return;
+    }
+
+    const text = keyName[keyName.length - 1];
```
urish, line 208: hmm doesn't `keyName[keyName.length - 1]` equal `key.toLowerCase()`? 

Also, with the current code, the following happens:
1. If you click on "A" with the mouse, the event is emitted with `{ detail: { key: "A" } }`
2. If you press "A" on the keyboard,  the event is emitted with `{ detail: { key: "a" } }` (note the difference)
3. If you press some other key, that does not exist on the keypad, e.g. the End key, an event is emitted with the last letter of the key name, i.e.  `{ detail: { key: "a" } }`

You can you the "EventLogger" panel of storybook to view the events that are sent for each keypress

src/membrane-keypad-element.ts
karinchechik: That's right. I tracked the difference between the numbers and the signs with this variable (8 vs * and 3 vs #). Now I use e.key and it is no longer needed. But I have a new problem now. Without this part, if I press shift+8 -> the * button seems pressed. Then if I release the shift before the 8, I get a press and release of 8 and the * is still shown as pressed. That also happens in the opposite way of pressing 8, then press shift and release 8 before the shift. The 8 still has the pressed class. The only way I can think of is checking the pressedKeys var: check for each case of 3,8,# and * with shift on keyStrokeDown and keyStrokeUp. Is this solution too hard-coded in your opinion?

PR #25, opened 2020-04-24
feat: add buzzer element

gfeun:
Hey Uri,

This is my attempt to implement a simple buzzer web component.
I followed your video tutorial to design it with Inkscape.

I wanted the component to display music notes when in use so i implemented a css animation when the attribute "value" is true (similar to the led element on/off). It uses a [MIT licensed music note icon](https://commons.wikimedia.org/wiki/Category:Open_Iconic#/media/File:Open_Iconic_musical_note.svg)

This is very raw since i'm not experienced with css. I am not aware of css animations browser compatibility so this could be a problem too.

### DIFF
--- a/src/buzzer-element.ts
+++ b/src/buzzer-element.ts
@@ -78,41 +78,34 @@ export class BuzzerElement extends LitElement {
 
         <svg
           width="17mm"
-          height="19.975mm"
+          height="20mm"
           version="1.1"
-          viewBox="0 0 17 19.975"
+          viewBox="0 0 17 20"
           xmlns="http://www.w3.org/2000/svg"
         >
-          <path d="m7.925 16.975v3" fill="none" stroke="#000" stroke-width=".5" />
-          <path d="m8.925 16.975v3" fill="#f00" stroke="#f00" stroke-width=".5" />
+          <path d="m8 16.5v3.5" fill="none" stroke="#000" stroke-width=".5" />
+          <path d="m9 16.5v3.5" fill="#f00" stroke="#f00" stroke-width=".5" />
           <g stroke="#000">
             <g>
-              <ellipse
+              <ellipse cx="8.5" cy="8.5" rx="8.15" ry="8.15" fill="#1a1a1a" stroke-width=".7" />
+              <circle
                 cx="8.5"
                 cy="8.5"
-                rx="8.1469"
-                ry="8.1469"
-                fill="#1a1a1a"
-                stroke-width=".70623"
-              />
-              <circle
-                cx="8.425"
-                cy="8.575"
-                r="6.2309"
+                r="6.3472"
                 fill="none"
                 stroke-width=".3"
                 style="paint-order:normal"
               />
               <circle
-                cx="8.425"
-                cy="8.575"
-                r="4.3137"
+                cx="8.5"
+                cy="8.5"
+                r="4.3488"
                 fill="none"
                 stroke-width=".3"
                 style="paint-order:normal"
               />
             </g>
-            <circle cx="8.425" cy="8.575" r="1.3744" fill="#ccc" stroke-width=".25129" />
+            <circle cx="8.5" cy="8.5" r="1.3744" fill="#ccc" stroke-width=".25" />
           </g>
         </svg>
       </div>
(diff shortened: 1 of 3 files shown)


urish:
This is lovely @gfeun!

buzzer

What other names can we have for the attribute that controls the music animation?

urish:
Also, I'm thinking it'd be cool to have two short wires or leads, possibly red and black, come out of the speaker body. When you search google images for "arduino speaker buzzer", most models that look like this one have red/blue wires coming out of them. This is a bit similar to the LED elements, which also have short leads sticking out of the LED body

src/buzzer-element.ts
```diff
@@ -0,0 +1,95 @@
...
+          />
+        </svg>
+        <svg
+          width="17mm"
+          height="17mm"
+          version="1.1"
+          viewBox="0 0 17 17"
+          xmlns="http://www.w3.org/2000/svg"
+        >
+          <g transform="matrix(.84999 0 0 .84999 8.5115e-5 -235.45)">
```
urish, line 85: Can you please remove the transforms? If you need assistance with that, please let me know

src/buzzer-element.ts
```diff
@@ -0,0 +1,95 @@
...
+        <svg
+          class="music-note"
+          style="visibility: ${buzzerOn ? '' : 'hidden'}"
+          xmlns="http://www.w3.org/2000/svg"
+          width="8"
+          height="8"
+          viewBox="0 0 8 8"
+        >
+          <path
+            d="M8 0c-5 0-6 1-6 1v4.09c-.15-.05-.33-.09-.5-.09-.83 0-1.5.67-1.5 1.5s.67 1.5 1.5 1.5 1.5-.67 1.5-1.5v-3.97c.73-.23 1.99-.44 4-.5v2.06c-.15-.05-.33-.09-.5-.09-.83 0-1.5.67-1.5 1.5s.67 1.5 1.5 1.5 1.5-.67 1.5-1.5v-5.5z"
```
urish, line 75: Not a deal breaker, but it'd be nice to combine this with the second SVG element. 
The `<path>` could be animated using CSS, similar to how you animate the `<svg>` element itself.

src/buzzer-element.ts
gfeun: I removed a translate transform using the method you shown in your video tutorial but didn't succeed in removing this matrix transform.
This may have been a mishap from my part.
I'll try again.

src/buzzer-element.ts
gfeun: Yes I'll do that.

src/buzzer-element.ts
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Renders a piezo electric buzzer.
+ */
+@customElement('wokwi-buzzer')
+export class BuzzerElement extends LitElement {
+  /**
+   * Boolean representing if an electric signal is coming in the buzzer
+   * If true a music note will be displayed on top of the buzzer
+   */
+  @property() value = false;
```
LironHazan, line 12: This is really cool 😎 maybe change the prop name to "hasSignal" so it would be more informative?

gfeun: Done :+1:

src/buzzer-element.ts
gfeun: So I have tried and pushed. However i can't animate the note outside of the svg size. This means that i had to increase the svg height so that the animation fits inside it. This could lead to difficulties placing the buzzer in a webpage as there is blank area on top of it.

urish: Thanks for looking into that Glenn! Yes, that makes sense. It sounds like the original approach you took (with two SVG elements) makes more sense in this light.

src/buzzer-element.ts
gfeun: I managed to remove the transform

src/buzzer-element.ts
gfeun: I reverted to 2 svg

gfeun:
I tried to add more details to the buzzer design but the contrast seems to make them undiscernible.
What do you think ?
Is there anything left to do ?

urish:
I think it's great the way it is. 

Do you happen to have a twitter account, btw?

PR #26, opened 2020-04-25
Rotary dailer

LironHazan:


urish:
Thanks @LironHazan!

Some feedback:
1. When you click on a digit, the circle-hole of that digit should go all the way until the finger stop. This usually takes less than a second, and then when the user leaves the mouse (which is the equivalent of lifting the finger), it goes back, a bit more slowly. You can get a better idea of the timings from this video:
https://youtu.be/v2oboe6gras?t=11
2. The numbers shouldn't be user selectable
3. What does the `value` knob do? if we don't need it can can remove it
4. Can you add keyboard support as well? 
5. Please add event listeners to the story, so we can see the events that are generated from the rotary dialer in storybook

LironHazan:
@urish Thanks for the feedback!! :)
Yes to everything! + the "value" is a generated leftover I forgot to remove :)

src/rotary-dialer-element.ts
```diff
@@ -0,0 +1,95 @@
...
+  removeDialerAnim() {
+    const dialerAnimClass = 'dialer-anim';
+    const slots = this.shadowRoot?.querySelector('#slots');
+    if (slots?.classList.contains(dialerAnimClass)) {
+      slots?.classList.remove(dialerAnimClass);
+    }
+  }
+
+  down(num: number) {
+    this.removeDialerAnim();
```
LironHazan, line 43: reminder --> maybe use .toggle() or toggleAttribute instead

talofer99:
Hi @LironHazan 

@urish sent me this element to review. and It looks super cool 

I got some feedback if that's ok 

1. It will be cool if the dial will have to be dragged by the mouse and not "auto dial" - giving some what of the original user experience
2. The dial should not go past the "finger blocker" in the current implement it looks like it gets all the way to the center of the dialed number,  its not possible if you had a finger there - hope I'm some what clear. 
3. the return looks as if it first "jumps" a bit back and then slowly move into place. It almost looks like it first flips the image horizontally.

LironHazan:
@talofer99 Hey I appreciate your input!  I'm always into the challenges :)  
I actually thought of the drugging functionality maybe as phase 2 (in general we need a dragging ability and I rather do something reusable).
 Regarding the animation behaviour you're right, the trick I'm doing is to register to the first animation end state and attach a new animation that does the reveres rotation, it's not so smooth and maybe the curve shouldn't be "ease out" as well, I'll tweak that. 
Regarding the dial itself, I'll read your comment again soon and will try to get it to act/feel as real as possible :)

src/rotary-dialer-element.ts
```diff
@@ -0,0 +1,280 @@
+import { css, customElement, html, LitElement, property } from 'lit-element';
+
+enum Digit {
```
urish, line 3: How about just narrowing the `number` type? e.g.

`type Digit = 0 | 1 |  2 | 3 | 4 | 5 | 6 | 7 | 8 | 9`

src/rotary-dialer-element.ts
```diff
@@ -0,0 +1,280 @@
...
+    this.handleRotation(digit);
+  }
+
+  private removeDialerAnim() {
+    const slots = this.shadowRoot?.querySelector('#slots');
+    slots?.classList.remove('dialer-anim');
+    slots?.removeAttribute('style');
+  }
+
+  private down(digit: number) {
```
urish, line 196: Wouldn't it be better to use `digit: Digit` in all places to keep it consistent?

src/rotary-dialer-element.ts
```diff
@@ -0,0 +1,280 @@
...
+            <text @mousedown=${() => this.down(3)} x="168" y="47">3</text>
+            <text @mousedown=${() => this.down(2)} x="210" y="79">2</text>
+            <text @mousedown=${() => this.down(1)} x="230" y="126">1</text>
+          </g>
+          <path
+            d="M182.738529,211.096297 L177.320119,238.659185 L174.670528,252.137377 L188.487742,252.137377 L182.738529,211.096297 Z"
+            id="Triangle"
+            stroke="#979797"
+            fill="#D8D8D8"
+            transform="translate(181.562666, 230.360231) rotate(-22.000000) translate(-181.562666, -230.360231) "
```
urish, line 273: Can you please remove the transform?

src/rotary-dialer-element.ts
```diff
@@ -0,0 +1,280 @@
...
+    return html`
+      <input
+        tabindex="0"
+        type="number"
+        class="hide-input"
+        value="${this.value}"
+        @input="${this.onValueChange}"
+      />
+      <svg width="266" height="266" xmlns="http://www.w3.org/2000/svg">
+        <g fill="none" fill-rule="evenodd">
```
urish, line 243: I think this outer `g` can be removed (you can move the `fill="none"` up to the `svg` element, if needed

src/rotary-dialer-element.ts
```diff
@@ -0,0 +1,280 @@
...
+  private onValueChange(event: KeyboardEvent) {
+    const target = event.target as HTMLInputElement;
+    const lastValue = target.value;
+    target.value = '';
+    const digit = parseInt(lastValue);
+    this.dial(digit);
+  }
+
+  private handleRotation(digit: number) {
+    const slots = this.shadowRoot?.querySelector('#slots') as SVGAElement;
```
urish, line 222: It's actually `SVGPathElement` and not [`SVGAElement`](https://developer.mozilla.org/en-US/docs/Web/API/SVGAElement)

src/rotary-dialer-element.ts
```diff
@@ -0,0 +1,280 @@
...
+    slots?.classList.remove('dialer-anim');
+    slots?.removeAttribute('style');
+  }
+
+  private down(digit: number) {
+    this.dial(digit);
+  }
+
+  /**
+   * Exposed lazy dial by applying dial fn with a given arg of num from 0-9
```
urish, line 201: Let's avoid abbreviations when writing doc strings, e.g. use "function" and "argument" instead of "fn" and "arg"

src/rotary-dialer-element.ts
```diff
@@ -0,0 +1,280 @@
...
+
+enum Direction {
+  normal = 'normal',
+  reverse = 'reverse',
+}
+@customElement('wokwi-rotary-dialer')
+export class RotaryDialerElement extends LitElement {
+  @property() value: number | '' = 0;
+
+  // todo: the degrees increases in ~40 when counting up the digits, would do that reusable if I could tweak the @keyframe from js
```
urish, line 24: Do you need any help with that?

src/rotary-dialer-element.ts
LironHazan: sure :) I know enums are heavy but I still tend to use them lol

LironHazan: know I see I did it for the pattern matching (in the switch case) I'll try to tweak that if it will stay readable

urish:
Here we go - GIF where you can see the animation jumping just before it reverses:

phone-dialer-animation-jump

src/rotary-dialer-element.ts
LironHazan: definitely :)

src/rotary-dialer-element.ts
LironHazan: I actually found how to do that by insert a css rules but I don't think that's a "way to go" what do you think? see this as a reference --> https://stackoverflow.com/questions/18481550/how-to-dynamically-create-keyframe-css-animations

src/rotary-dialer-element.ts
LironHazan: will do!

src/rotary-dialer-element.ts
LironHazan: removing it will change the finger stop triangle position I need it a bit rotated :)

src/rotary-dialer-element.ts
LironHazan: removed

src/rotary-dialer-element.ts
```diff
@@ -0,0 +1,297 @@
...
+        new CustomEvent('dial', {
+          detail: { digit },
+        })
+      );
+    };
+    const rAF = requestAnimationFrame(() => {
+      slots?.classList.add('dialer-anim');
+      // When you click on a digit, the circle-hole of that digit
+      // should go all the way until the finger stop.
+      RotaryDialerElement.matchDigit(slots, digit);
```
urish, line 242: Why make `matchDigit()` static? It requires repeating the class name whenever you use it...

src/rotary-dialer-element.ts
```diff
@@ -0,0 +1,297 @@
...
+            id="slots"
+            stroke="#fff"
+            fill-opacity="0.5"
+            fill="#D8D8D8"
+          ></path>
+        </g>
+        <circle id="container" fill-opacity=".5" fill="#070707" cx="132.5" cy="132.5" r="132.5" />
+        <g class="text" font-family="Monaco" font-size="21" fill="#FFF">
```
urish, line 275: This font is not a [safe web font](https://www.w3schools.com/cssref/css_websafe_fonts.asp), so let's provide a fallback for people who don't have this font (unless you are okay with the default fallback)

src/rotary-dialer-element.ts
```diff
@@ -0,0 +1,297 @@
...
+          <text @mouseup=${() => this.up(6)} x="35" y="97">6</text>
+          <text @mouseup=${() => this.up(5)} x="72" y="58">5</text>
+          <text @mouseup=${() => this.up(4)} x="117" y="41">4</text>
+          <text @mouseup=${() => this.up(3)} x="168" y="47">3</text>
+          <text @mouseup=${() => this.up(2)} x="210" y="79">2</text>
+          <text @mouseup=${() => this.up(1)} x="230" y="126">1</text>
+        </g>
+        <path
+          d="M182.738529,211.096297 L177.320119,238.659185 L174.670528,252.137377 L188.487742,252.137377 L182.738529,211.096297 Z"
+          id="Triangle"
```
urish, line 289: Why is the `id` needed? Is it used somewhere?

src/rotary-dialer-element.ts
```diff
@@ -0,0 +1,297 @@
...
+          detail: { digit },
+        })
+      );
+    };
+    const rAF = requestAnimationFrame(() => {
+      slots?.classList.add('dialer-anim');
+      // When you click on a digit, the circle-hole of that digit
+      // should go all the way until the finger stop.
+      RotaryDialerElement.matchDigit(slots, digit);
+      cancelAnimationFrame(rAF);
```
urish, line 243: AFAIK you don't need to cancel it, because the callback for request animation frame is only called once

src/rotary-dialer-element.ts
```diff
@@ -0,0 +1,297 @@
...
+      }
+      .dialer-anim {
+        transform-origin: center;
+      }
+      @keyframes dialOne {
+        0% {
+          transform: rotate(0);
+        }
+        50% {
+          transform: rotate(56deg);
```
urish, line 38: Not sure if that would work, but perhaps you can store the target degree in a CSS variable and in this way reuse a single animation for all the digits?

Another way I'm thinking off would be to replace the animations with transitions, and then when a digit is dialed, you apply a class to start the transition on the `transform` property. When it ends you remove that class. Not sure if it'd be better than the current approach, just an idea...

src/rotary-dialer-element.ts
```diff
@@ -0,0 +1,297 @@
...
+  private onValueChange(event: KeyboardEvent) {
+    const target = event.target as HTMLInputElement;
+    const digit = parseInt(target.value) as Digit;
+    this.dial(digit);
+    target.value = '';
+  }
+
+  private handleRotation(digit: Digit) {
+    const slots = this.shadowRoot?.querySelector('#slots') as SVGPathElement;
+    slots.onanimationend = () => {
```
urish, line 231: You can use a lit element event listener instead of manually attaching the event listener every time

src/rotary-dialer-element.ts
```diff
@@ -0,0 +1,297 @@
...
+  private static matchDigit(
+    slots: SVGPathElement,
+    digit: Digit,
+    direction: Direction = Direction.normal,
+    duration = '1000ms',
+    timingFn = 'ease-out'
+  ) {
+    switch (digit) {
+      case 0:
+        slots?.setAttribute('style', `animation: dialZero ${duration} 1 ${direction} ${timingFn};`);
```
urish, line 156: There is a lot of repetition here, and if I understand correctly, you can replace the switch statement with an object that will map from digit to its name, or rename the animations to dial0, dial1, etc. so you can use the same line for all the animations.

Also, you can probably avoid the direct DOM manipulating by changing some instance variable here, and then using this variable in the `render()` function so set the respective style on the slots element.

src/rotary-dialer-element.ts
urish: See my suggestions below - perhaps using a CSS variable, or using a CSS `transition` instead will help here

src/rotary-dialer-element.ts
LironHazan: My code inspector (webstorm) always suggesting a potential pure util function to be static, I actually like it, we don't have any reference or use to the instance inside i, we're not gonna need it on the prototype as well :)

src/rotary-dialer-element.ts
LironHazan: sorry I missed that again! will guard

src/rotary-dialer-element.ts
LironHazan: nope! leftovers..  thanks!

src/rotary-dialer-element.ts
LironHazan: yes you're right.. that's a habit I have in clearing and unsubscribing and destroying :) will remove!

src/rotary-dialer-element.ts
LironHazan: css variable won't work for the @keyframe AFAIK I could try insert a new rule (hardcore), regarding transitions if I'll use that than I could use css var for the transform - let me try that direction :) I like it

src/rotary-dialer-element.ts
LironHazan: yes that would be better!

src/rotary-dialer-element.ts
LironHazan: Thats a very good feedback, if there was a certain math pattern that would be best cause than I was just iterating the numbers and increasing the degrees according to a const or a calc result but since I couldn't find one I did the switch-case.
 
At first when I had the reverse I passed different inputs to the functions but now it's definitely not DRY and you're right I can use an object mapper, I'll think I'll try to switch to transition with classes instead of animation and style - that will change the current implementation, if I'll eventually stay with animation keyframes I'll change that function according to your comment :)

src/rotary-dialer-element.ts
LironHazan: @urish removing the transition class onanimationend didn't work :/ I'll DRY the anaimation code

urish: When working with CSS transitions, it seems like the event you want to listen for is [`transitionend`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/transitionend_event). Have you tried that?

LironHazan: hey just noticed, no I didnt try that! I naively assumed that it uses the onanimationend, glad to learn something new, will give. it a try!

PR #28, opened 2020-05-28
fix(dialer): firing each dial event once

LironHazan:


LironHazan:
@urish thanks for the detailed issue! 🚀

src/rotary-dailer-element.stories.ts
```diff
@@ -7,4 +7,6 @@ import { logEvent } from 'storybook-events-logger';
 storiesOf('Rotary Dialer', module)
   .addParameters({ component: 'wokwi-rotary-dialer' })
   .addDecorator(withKnobs)
-  .add('Default', () => html` <wokwi-rotary-dialer @dial=${logEvent}></wokwi-rotary-dialer> `);
+  .add('Start', () => html` <wokwi-rotary-dialer @dial-start=${logEvent}></wokwi-rotary-dialer> `)
```
urish, line 10: Why not keep a single story with all three events?
This way it's easier to see the temporal relation between the events (i.e. their order).

Usually different stories are used to display different visual style options of the element

src/rotary-dialer-element.ts
```diff
@@ -91,9 +104,12 @@ export class RotaryDialerElement extends LitElement {
           <circle stroke="#fff" stroke-width="2" fill="#D8D8D8" cx="133" cy="133" r="72" />
           <path
             class=${classMap(this.classes)}
+            @transitionstart="${() => {
+              this.dialStartCount = this.fireOnce(this.dialStartCount, 'dial');
+            }}"
             @transitionend="${() => {
               this.removeDialerAnim();
-              this.dispatchEvent(new CustomEvent('dial', { detail: { digit: this.digit } }));
+              this.dialEndCount = this.fireOnce(this.dialEndCount, 'dial-end');
```
urish, line 112: How about looking at the CSS class to determine whether it's the middle or the end of the dial event?

src/rotary-dailer-element.stories.ts
LironHazan: noted! I'lll change

Issue #8, opened 2020-03-27
Add Arduino Nano Element


urish:
The board needs to include at least the male pin headers, the Mini-USB connector, 4 LEDs (TX, RX, Power, L), the microcontroller itself, and the ICSP programmer pins (can be seen on the right side of the board). Nice to have elements: reset button (which should also emit reset-press/reset-release events, similar to the [pushbutton element](https://github.com/wokwi/wokwi-elements/blob/master/src/pushbutton-element.ts).

As a reference, check out the implementation of the Arduino Uno Element. You can reuse most of the part, especially the pattern for drawing the pin headers, USB/power jacks, LEDs, and the ICSP programmer pins.

Reference images:

[![Arduino Nano](https://user-images.githubusercontent.com/892318/77750844-91357a00-7035-11ea-9525-742fdddc0054.png)](https://www.piqsels.com/en/public-domain-photo-zadii)

License: CC0

[![Arduino Nano](https://upload.wikimedia.org/wikipedia/commons/7/78/Arduino_nano.jpg)](https://upload.wikimedia.org/wikipedia/commons/7/78/Arduino_nano.jpg)

License: CC-BY-SA

Mechanical drawings:
[![Ardunio Nano Machanical Drawing](https://nulliusinverba915909567.files.wordpress.com/2018/11/arduino-nano-v30-atmega328-blue-4.jpg)](https://nulliusinverba.blog/tag/mechanical-drawing/)

[![Arduino Naon Mechanical Drawing](https://components101.com/sites/default/files/inline-images/Arduino-Nano-Dimensions.png)](https://components101.com/microcontrollers/arduino-nano)

sutaburosu:
@urish I'm a total noob to Node, Typescript, SVG, graphic design and webdev in general.  I learned a huge amount from your tutorial videos and yesterday's commit adding the Mega.  There's no rush for this, right?  I think I can do this,  but it will take me a while.  I may need help later getting the reset button to work.

My only progress so far is this pattern for unsoldered pins.  Good enough?
![image](https://user-images.githubusercontent.com/20606/92040131-2c154b80-ed6e-11ea-86bc-8914b43652a9.png)

```
import { svg } from 'lit-element';

export const pinsPlatedThroughHolePattern = svg`
  <defs>
    <filter id="solderPlate" style="color-interpolation-filters:sRGB;">
      <feTurbulence result="r0" type="fractalNoise" baseFrequency="1" numOctaves="3"/>
[...]

PR #43, opened 2020-09-03
feat(nano): WIP. Done except for reset button (#8)

sutaburosu:
Regarding the reset button, am I supposed to just duplicate what I need from `pushbutton-element.ts` into the Nano element ?  Is there a better, DRYer way?

The x,y locations in `pinInfo[]` have **not** been verified to be correct,  Getting avr8js running so I can check them is my next task.

image

### DIFF
--- a/src/arduino-nano-element.stories.ts
+++ b/src/arduino-nano-element.stories.ts
@@ -1,0 +1,21 @@
+import { boolean, withKnobs, button } from '@storybook/addon-knobs';
+import { action } from '@storybook/addon-actions';
+import { storiesOf } from '@storybook/web-components';
+import { html } from 'lit-html';
+import './arduino-nano-element';
+
+storiesOf('Arduino Nano', module)
+  .addParameters({ component: 'wokwi-arduino-nano' })
+  .addDecorator(withKnobs)
+  .add(
+    'Nano',
+    () => html`
+      <wokwi-arduino-nano
+        .led13=${boolean('LED 13', false)}
+        .ledTX=${boolean('TX LED', false)}
+        .ledRX=${boolean('RX LED', false)}
+        .ledPower=${boolean('Power LED', true)}
+        .resetButton=${boolean('Reset button', false)}
+      ></wokwi-arduino-nano>
+    `
+  );
--- a/src/arduino-nano-element.stories.ts
+++ b/src/arduino-nano-element.stories.ts
@@ -1,7 +1,7 @@
-import { boolean, withKnobs, button } from '@storybook/addon-knobs';
-import { action } from '@storybook/addon-actions';
+import { boolean, withKnobs } from '@storybook/addon-knobs';
 import { storiesOf } from '@storybook/web-components';
 import { html } from 'lit-html';
+import { logEvent } from 'storybook-events-logger';
 import './arduino-nano-element';
 
 storiesOf('Arduino Nano', module)
@@ -16,6 +16,8 @@ storiesOf('Arduino Nano', module)
         .ledRX=${boolean('RX LED', false)}
         .ledPower=${boolean('Power LED', true)}
         .resetButton=${boolean('Reset button', false)}
+        @button-press=${logEvent}
+        @button-release=${logEvent}
       ></wokwi-arduino-nano>
     `
   );
--- a/src/arduino-nano-element.stories.ts
+++ b/src/arduino-nano-element.stories.ts
@@ -15,7 +15,6 @@ storiesOf('Arduino Nano', module)
         .ledTX=${boolean('TX LED', false)}
         .ledRX=${boolean('RX LED', false)}
         .ledPower=${boolean('Power LED', true)}
-        .resetPressed=${boolean('Reset button', false)}
         @button-press=${logEvent}
         @button-release=${logEvent}
       ></wokwi-arduino-nano>
--- a/src/arduino-nano-element.stories.ts
+++ b/src/arduino-nano-element.stories.ts
@@ -15,7 +15,7 @@ storiesOf('Arduino Nano', module)
         .ledTX=${boolean('TX LED', false)}
         .ledRX=${boolean('RX LED', false)}
         .ledPower=${boolean('Power LED', true)}
-        .resetButton=${boolean('Reset button', false)}
+        .resetPressed=${boolean('Reset button', false)}
         @button-press=${logEvent}
         @button-release=${logEvent}
       ></wokwi-arduino-nano>
--- a/src/arduino-nano-element.ts
+++ b/src/arduino-nano-element.ts
@@ -1,13 +1,15 @@
 import { customElement, html, LitElement, property, svg } from 'lit-element';
 import { analog, ElementPin, i2c, spi, usart } from './pin';
 
+const triggerKeys = [32, 13]; // space and enter
+
 @customElement('wokwi-arduino-nano')
 export class ArduinoNanoElement extends LitElement {
   @property() led13 = false;
   @property() ledRX = false;
   @property() ledTX = false;
   @property() ledPower = false;
-  @property() resetButton = false;
+  @property() resetPressed = false;
 
   readonly pinInfo: ElementPin[] = [
     { name: '12', x: 19.7, y: 4.8, signals: [spi('MISO')] },
@@ -49,8 +51,19 @@ export class ArduinoNanoElement extends LitElement {
   ];
 
   render() {
-    const { ledPower, led13, ledRX, ledTX, resetButton } = this;
+    const { ledPower, led13, ledRX, ledTX, resetPressed } = this;
     return html`
+      <style>
+        text {
+          user-select: none;
+        }
+        circle[tabindex]:hover,
+        circle[tabindex]:focus {
+          stroke: white;
+          outline: none;
+        }
+      </style>
+
       <svg
         width="44.9mm"
         height="17.8mm"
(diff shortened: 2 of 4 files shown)


urish:
Oh wow, first of all - thanks a million :-)

I hope to get a look at the files soon. Regarding the reset button - that's a good question. The easiest solution would probably to add a click listener on the group that contains the button. If we want to support keyboard interaction, a believe a good place to look for inspiration would be the implementation of the [keypad element](https://github.com/wokwi/wokwi-elements/blob/f698e0c8ee39debcfaf6b59f5508cdb05d786c92/src/membrane-keypad-element.ts#L71), where we use `tabindex` to make it focusable, and then listen for mouse/touch/keyboard event.

However, I think that in case of the reset button, we don't actually need all the complexity - it'd probably suffice just to emit a 'click' event, since we only care about the face it was clicked...

urish:
p.s. it looks gorgeous, and the code also seems very tidy! great job :-)

sutaburosu:
Thanks.  I've surprised myself.  This time yesterday I had never used Inkscape, and I hadn't even heard of React, etc.

I've enjoyed myself a great deal,  but now the difficult bit starts:  trying to learn more about how this works,  rather than simply mimicking what I've seen you do.

urish:
I'm here to help :-)

Regarding pinInfo, this is what I did with the MEGA (I should have documented it, but I was rushing to get it out...):

I wrapped the `<svg>` element with another `<div>`, and then created a second `<svg>` inside it to show me the pin positions. This is what it looked like:

before: 
```typescript
render() {
  return `<svg...>blabla</svg>`;
}
```

After:
```typescript
render() {
  return `
    <div style="position: relative">
      <svg...>blabla</svg>

      <svg style="position: absolute; top: 0; left: 0" width="100%" height="100%" fill="red">
        ${this.pinInfo.map(pin => svg`<circle cx=${pin.x} cy=${pin.y} r=2 />`)}
      </svg>
    </div>
  `;
}
```

and then remove the extra code once you are done.

Ideally, I'd love to have some Storybook addon/integration that will do it automatically (so we can click on a button and see all the pinInfos), but for now this is the method that I've been using as a workaround...

sutaburosu:
Thanks for the tip. If I add the map() within the existing SVG, everything aligns.  Then, after using a \<div\> with 2 \<svg\> like you said it shrunk to like 25% the size.  Multiplying my mm coords by 1 / 0.254 made it fit, so I deduced that 2nd \<svg> is using 0.1in units.

But then I looked at the Mega pins.  They seem be in units of 1/100 inch.  Then it occurred to me that monitors are frequently 96 DPI and that is very close to 0.01".  Is pinMap actually in 1/96th inch pixels?  Does that work OK on high-DPI screens?

I also have a problem where all but the last \<tspan> in a \<text> is misaligned by x-=0.1 inch.  I'm starting to think I may have found a bug in something.

I'll spend some time on the reset button logic later today.

sutaburosu:
Thanks for the links. That aspect now makes a lot more sense, and I see that much of what I wrote is actually CSS styles rather than SVG. That's good to know.

I've learned what causes my [centred text positioning problems](https://stackoverflow.com/a/23272711). I can't see a way to use `text-anchor="middle"` at all with \<tspan> without having the layout broken by the Prettier git commit hook. To work around this, I positioned one \<text> per item, without the useful `dx` automatic relative positioning of \<tspan>.

I've done what I can with the reset button by dispatching a `CustomEvent`. similar to what's in the membrane.  I'm struggling to see where this event might be received so I can make the AVR reset.

src/arduino-nano-element.ts
```diff
@@ -0,0 +1,277 @@
+import { customElement, html, LitElement, property, svg } from 'lit-element';
+import { analog, ElementPin, i2c, spi, usart } from './pin';
+
+@customElement('wokwi-arduino-nano')
+export class ArduinoNanoElement extends LitElement {
+  @property() led13 = false;
+  @property() ledRX = false;
+  @property() ledTX = false;
+  @property() ledPower = false;
+  @property() resetButton = false;
```
urish, line 10: I think it'll make more sense to call it `resetPressed` (since it determines whether to draw the reset button in the pressed state or not, not whether to display it / turn it on/off)

src/arduino-nano-element.ts
```diff
@@ -0,0 +1,277 @@
...
+          @mouseup=${() => this.up()}
+          @mouseleave=${() => this.up()}
+          @touchstart=${() => this.down()}
+          @touchend=${() => this.up()}
+        >
+          <rect x="24.3" y="6.3" width="1" height="4.8" filter="url(#solderPlate)" fill="#333" />
+          <rect x="23.54" y="6.8" width="2.54" height="3.8" fill="#ccc" stroke="#888" />
+          <ellipse cx="24.8" cy="8.7" rx="1" ry="1" fill="#eeb" stroke="#777" />
+          ${resetButton &&
+          svg`<circle cx="24.8" cy="8.7" r="1.5" fill="#f66" filter="url(#ledFilter)" />`}
```
urish, line 254: It makes more sense to change the color of the ellipse above the a darken shade when the button is pressed (we could also use a gradient to make the graphics a bit more realistic, similar to how the pushbutton works). 

Right now it seems like the button is illuminated whenever you press it.

src/arduino-nano-element.ts
```diff
@@ -0,0 +1,277 @@
...
+          svg`<circle cx="1.3" cy=".55" r="1.3" fill="#80ff80" filter="url(#ledFilter)" />`}
+        </g>
+        <g transform="translate(27.7 11)">
+          <use xlink:href="#led-body" />
+          ${led13 &&
+          svg`<circle cx="1.3" cy=".55" r="1.3" fill="#ffff80" filter="url(#ledFilter)" />`}
+        </g>
+
+        <!-- reset button -->
+        <g
```
urish, line 240: how about adding a `tabindex="0"` attribute, so that the reset will be focusable by keyboard (using the tab key?)

urish (changes requested):

urish:
So there are some ways to work around prettier (such as wrapping the element that you don't want to be formatted with 

```
   ${svg` <text>dont format me!  </text>  `}
```

and there's also prettier-ignore (but it doesn't play as nicely with lit-element code), but in general, I like to stick with prettier whenever possible. 

I've left some comments on the PR, and I hope that you will be able to look into them.

Re: the reset - it's actually not implemented yet at the simulator level. Right now the best way to "reset" is to restart the simulation altogether. Once the new nano lands, I'll start looking into restarting just the simulated MCU. If I'm not wrong, most peripherals already implement a `reset()` method, so it could be as simple as calling that method on all the peripherals and the CPU instance.

src/arduino-nano-element.ts
sutaburosu: Good point. Naming things is hard. Fixed.

src/arduino-nano-element.ts
sutaburosu: I'm happy to make those changes, but perhaps I should first explain why I did it this way.  At default zoom, the Nano is very small and the button is *tiny*.  I figured just reversing a gradient wouldn't be visible.  I have an MCU (a Teensy I think) with a red LED inside a similar tiny button, so it's not an invention of mine, just a little artistic license. :)

src/arduino-nano-element.ts
sutaburosu: I've taken a shot at this in the latest commit.  Much googling was required.  I'm still not 100% happy.  I would like the button to lose focus on mouseleave, but I'm struggling to make that happen.

I've left the LED in pending your feedback re the LED/gradient.

src/arduino-nano-element.ts
urish: Yes, a gradient may be to subtle in default zoom. The LED still seems a bit weird for me, how about just a darker shade for the round cap? e.g. #333

src/arduino-nano-element.ts
```diff
@@ -0,0 +1,308 @@
...
+            r="1"
+            fill="#eeb"
+            stroke="#777"
+            tabindex="0"
+            @mousedown=${() => this.down()}
+            @touchstart=${() => this.down()}
+            @mouseup=${() => this.up()}
+            @mouseleave=${() => this.leave()}
+            @touchend=${() => this.up()}
+            @keydown=${(e: KeyboardEvent) => triggerKeys.includes(e.keyCode) && this.down()}
```
urish, line 270: `e.keyCode` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) (I will also open an issue to remove it from the other elements), we can use [`e.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) or [`e.code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) instead

src/arduino-nano-element.stories.ts
```diff
@@ -0,0 +1,23 @@
...
+  .addDecorator(withKnobs)
+  .add(
+    'Nano',
+    () => html`
+      <wokwi-arduino-nano
+        .led13=${boolean('LED 13', false)}
+        .ledTX=${boolean('TX LED', false)}
+        .ledRX=${boolean('RX LED', false)}
+        .ledPower=${boolean('Power LED', true)}
+        .resetPressed=${boolean('Reset button', false)}
```
urish, line 18: We can leave this property (`resetPressed`) out of the story (it's not something the users of this element should usually change)

urish (changes requested):
Cool! Some more small things, and we're good to go!

src/arduino-nano-element.ts
urish: Shared my feedback. What was the rationale behind losing focus on mouseleave?

sutaburosu: Thanks.  Losing focus on mouseleave is purely for appearance.  It didn't feel right that the button looked different before/after clicking it, until it lost focus somehow.

sutaburosu:
I've done those two small changes.  I initially used CSS `active` to restyle the circle when pressed, but of course that doesn't work with the keyboard so instead I manually restyle it in `up()` and `down()`

urish:
🎉

PR #45, opened 2020-09-07
Remove usage of depracated KeyboardEvent.keyCode

### LINKED ISSUE #44: Remove usage of depracated KeyboardEvent.keyCode
The following components use `KeyboardEvent.keyCode`, which is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode): 

* [pushbutton-element](https://github.com/wokwi/wokwi-elements/blob/f698e0c8ee39debcfaf6b59f5508cdb05d786c92/src/pushbutton-element.ts#L49)
* [membrane-keypad-element](https://github.com/wokwi/wokwi-elements/blob/f698e0c8ee39debcfaf6b59f5508cdb05d786c92/src/membrane-keypad-element.ts#L84)

We can switch to [`KeyboardEvent.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) or [`KeyboardEvent.code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) instead.

takuma-watanabe:
This PR resolve #44 .

### DIFF
--- a/src/membrane-keypad-element.ts
+++ b/src/membrane-keypad-element.ts
@@ -1,8 +1,7 @@
 import { customElement, html, LitElement, property, svg } from 'lit-element';
 import { pinsFemalePattern } from './patterns/pins-female';
 import { ElementPin } from './pin';
-
-const SPACE_KEYS = [' ', 'Spacebar'];
+import { SPACE_KEYS } from './utils/keys';
 
 const rowPositions = [10.7, 25, 39.3, 53.6];
 const columnPositions = [7, 22, 37, 52];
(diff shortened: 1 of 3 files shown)


urish:
First of all, thank you! 

I'm thinking it might make even more sense to move the `SPACE_KEYS` array into a common file (e.g. `src/utils/keys.ts` or something similar), instead of redefining it in each element that needs it. WDYT?

takuma-watanabe:
Thanks for your review!
I will try to move `SPACE_KEYS` array to common file.

takuma-watanabe:
@urish I fixed. Please review again.

sutaburosu:
@takuma-watanabe there's also `triggerKeys` in the Nano element that could be adapted to use `utils\keys.ts`.  I included `Enter` in that list because I found pressing Space could also scroll the window in some circumstances.

urish:
good point @sutaburosu, I'd love to see this in a subsequent PR

sutaburosu:
I'll do that now. :)

takuma-watanabe:
@sutaburosu I couldn't notice that point. Thank you.

PR #73, opened 2021-05-28
Update ntc-temperature-sensor-element.ts

matititam:
removed redundant text

### DIFF
--- a/src/ntc-temperature-sensor-element.ts
+++ b/src/ntc-temperature-sensor-element.ts
@@ -111,13 +111,13 @@ export class NTCTemperatureSensorElement extends LitElement {
           stroke-linejoin="miter"
           stroke-width=".65px"
         />
-          <text fill="#fffefe" font-family="sans-serif" transform="rotate(-90)">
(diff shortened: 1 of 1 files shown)


urish (changes requested):
we only have one <text> element now, so can you please also remove this <g> and move the `fill` and `font-family` directly into the `<text>` element below?

src/ntc-temperature-sensor-element.ts
```diff
@@ -111,15 +111,13 @@ export class NTCTemperatureSensorElement extends LitElement {
           stroke-linejoin="miter"
           stroke-width=".65px"
         />
-        <g fill="#fffefe" font-family="sans-serif">
```
urish, line 114: Great! Now just reformat the file with prettier to keep the indentation correct ( F1 then "Format Document" inside gitpod / vscode should probably do the trick)

urish (changes requested):

src/ntc-temperature-sensor-element.ts
urish: yes, thanks!

PR #75, opened 2021-05-31
Update flame-sensor-element.ts

matititam:
added 4th element and 4th pin
updated label and one white rectangle around 4 pins


urish:
Preview: https://deploy-preview-75--wokwi-elements.netlify.app/?path=/story/flame-sensor--default

seems like the forth pin isn't rendering?

Also, it's possible to optimize the SVG by defining the `<path>` once inside a `<defs>` section, modifying the starting coordinates to be `m0 0`, then making several copies it with the `<use>`  element. You can see in example in the [servo element](https://github.com/wokwi/wokwi-elements/blob/master/src/servo-element.ts#L67-L69)

PR #77, opened 2021-06-02

arcostasi:


urish:
Thanks!

Preview URL: https://deploy-preview-77--wokwi-elements.netlify.app/?path=/story/franzininho--default

The board looks great, but:

1. What's the license for the graphics? If it's MIT/BSD/CC-BY/compatible, then great, but we need to include the credits in the source file. For other licenses, we need to see if they can be included in the library.
2. The SVG contains many redundant elements. For instance, there's a bunch of empty `<text>` elements. Did you export it as optimized SVG from Inkscape? In general, we try to keep the elements small (< 10kb), so that they load and render fast. The current file is over 500kb...
3. The pin locations are off - note that there's a [`<wokwi-show-pins>`](https://github.com/wokwi/wokwi-elements/blob/master/CONTRIBUTING.md#debugging-pin-info) element that you can use for debugging.
4. For the LEDs, you can define two boolean properties: `ledPower` and `led1`, similar to what we have in the Arduino Nano

PR #81, opened 2021-06-17

arcostasi:


src/franzininho-element.ts
```diff
@@ -0,0 +1,455 @@
...
+    { name: '5V.1', x: 218.5, y: 32.9, signals: [{ type: 'power', signal: 'VCC', voltage: 5 }] },
+    { name: 'PB5', x: 218.5, y: 42.5, signals: [analog(0)] },
+    { name: 'PB4', x: 218.5, y: 52.2, signals: [analog(2), { type: 'pwm' }] },
+    { name: 'PB3', x: 218.5, y: 61.7, signals: [analog(3)] },
+    { name: 'PB2', x: 218.5, y: 71.2, signals: [spi('SCK'), i2c('SCL'), analog(1)] },
+    { name: 'PB1', x: 218.5, y: 80.9, signals: [spi('MISO'), { type: 'pwm' }] },
+    { name: 'PB0', x: 218.5, y: 90.5, signals: [spi('MOSI'), i2c('SDA'), { type: 'pwm' }] },
+    { name: 'VIN', x: 132.7, y: 8.1, signals: [{ type: 'power', signal: 'VCC' }] },
+    { name: 'GND.2', x: 142.9, y: 8.1, signals: [{ type: 'power', signal: 'GND' }] },
+    { name: '5V.1', x: 153, y: 8.1, signals: [{ type: 'power', signal: 'VCC', voltage: 5 }] },
```
urish, line 24: Duplicate pin name. Also, I think it may make more sense to call it "VCC.2" (and the other one VCC.1) as this is the text written next to it on the PCB

src/franzininho-element.stories.ts
```diff
@@ -0,0 +1,20 @@
+import { withKnobs, boolean } from '@storybook/addon-knobs';
+import { storiesOf } from '@storybook/web-components';
+import { html } from 'lit-html';
+import { action } from '@storybook/addon-actions';
+import './franzininho-element';
+
+storiesOf('Franzininho', module)
```
urish, line 7: Can you please change the story to the new storybook 6 format? 

You can find an example in the flame-sensor or slide-potentiometer element stories

urish (changes requested):
Thank you so much!

PR #82, opened 2021-06-17

arcostasi:


urish:
Thank you!

arcostasi:
Thanks Uri

urish:
Released as wokwi-elements 0.41.0

should be available in Wokwi.com soon

Issue #83, opened 2021-06-18
Simulation rendering in Safari


Koepel:
Firefox rendering of resistors:
![firefoxrender](https://user-images.githubusercontent.com/12096541/122611955-8aec4d00-d082-11eb-9847-3154807580e6.png)

Safari (11.1.2) rendering of resistors:
![safarirender](https://user-images.githubusercontent.com/12096541/122611988-9b042c80-d082-11eb-8c8c-1e038e038694.png)

Related to: https://github.com/wokwi/wokwi-features/issues/89

Additional information: The latest and obsolete Internet Explorer in Windows 10 does not render the circuit simulation and the PaleMoon browser shows a grey screen for the circuit simulation part. Every other browser in Windows 10, linux and Android seems to be okay.

sutaburosu:
![Safari dev console](https://user-images.githubusercontent.com/20606/122612578-47daab80-d07b-11eb-9e18-0004f85fc319.png) when loading [this sketch](https://wokwi.com/arduino/projects/300936732038136328).

PR #84, opened 2021-06-18
fix(resistor): Safari barfs on spaces in CSS properties

sutaburosu:
Should fix issue #83 .


urish:
Thank you so much Steve!

PR #90, opened 2021-06-29
fix(nano): I2C pins reversed

sutaburosu:
Apparently I reversed the two I2C pins.


urish:
Thanks! Fixed in 8efe7533e0064f01c92d479572c83a8a406edaa0

PR #114, opened 2021-12-29
feat: add flip functionality to the LED element

Guy-Rahamim:
LEDs can now be flipped using the "flipped" attribute in the diagram.json file.
Also added a "Flipped" story to the storybook.

### DIFF
--- a/src/led-element.ts
+++ b/src/led-element.ts
@@ -18,11 +18,17 @@ export class LEDElement extends LitElement {
   @property() color = 'red';
   @property() lightColor: string | null = null;
   @property() label = '';
+  @property({ type: Boolean }) flip = false;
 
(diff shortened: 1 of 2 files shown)


src/led-element.ts
```diff
@@ -18,11 +18,17 @@ export class LEDElement extends LitElement {
   @property() color = 'red';
   @property() lightColor: string | null = null;
   @property() label = '';
+  @property() flipped = false;
```
urish, line 21: Let's specify that it's a boolean input, like here: https://github.com/wokwi/wokwi-elements/blob/9790133c7ac9a2f77eb66a2beb2aa26ceadd1937/src/7segment-element.ts#L21

urish:
Thanks!

PR #115, opened 2022-01-07
Adds a PC Fan and Robot Arm element

yepher:
This creates a PC Fan and Robot Arm element

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -48,3 +48,5 @@ export { LedBarGraphElement } from './led-bar-graph-element';
 export { MicrosdCardElement } from './microsd-card-element';
 export { DipSwitch8Element } from './dip-switch-8-element';
 export { StepperMotorElement } from './stepper-motor-element';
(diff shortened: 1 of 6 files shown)


urish:
Thanks Chris!

I love it. So let's break it into two PRs: one for the fan and the other one for the robotic arm?

Also, regarding the robotic arm, this brings an interesting question: what should the bounding box be?

yepher:
I will break it up. And was wondering the same thing about the bounding box. Previously I took the full space the arm can move. Not sure if it can be dynamic or possibly draw outside it's box...

urish:
Thanks!

Dynamic - yes. Not sure if it'll work in Wokwi out of the box though - we can try and see if it breaks anything.

Draw out of the box - not in SVG. You can resize the SVG, though, and move it relative to the container element, so Wokwi still thinks the box (the containing elements) is smaller, and draw the marching ants around it. I'm not sure if this will lead to good UX, however?

Issue #101, opened 2021-08-02
HX710B Pressure Sensor


arcostasi:
This HX710B air pressure sensor module uses a high-precision AD sampling chip, adopts a 0-40KPa air pressure sensor, can connect a 2.5mm hose, can detect water level, and other air pressure.

![MPS20N0020D_Arduino_wiring](https://user-images.githubusercontent.com/5041497/127874649-2cf2a682-9a2b-4d34-9a06-52e27de304d0.png)

### Technical Specifications

- VIN: Module power supply – 3.3-5 V
- GND: Ground
- SLC: I2C Clock
- OUT: Digital output data
- Pressure: 0-40KPa
- Gain: 32 / 64 / 128
- Size: 19x18mm (2mm fixing hole)

### Pinouts

![MPS20N0040D-pin](https://user-images.githubusercontent.com/5041497/127874451-dec6dda2-bb72-466d-8419-83c8de5eec96.jpg)

### Documents

- [Pressure Sensor Guide](https://www.electroschematics.com/pressure-sensor-guide/)

### Code

```cpp
/* 
  Source: https://www.electroschematics.com/pressure-sensor-guide/
*/
#include "HX711.h"

HX711 scale;

void setup() {
  Serial.begin(9600);
  Serial.println("HX710B Demo with HX711 Library");
  Serial.println("Initializing the scale");

  // parameter "gain" is ommited; the default value 128 is used by the library
  // HX711.DOUT   - pin #A1
  // HX711.PD_SCK - pin #A0

  scale.begin(A1, A0);

  Serial.println("Before setting up the scale:");
  Serial.print("read: \t\t");
  Serial.println(scale.read());                  // print a raw reading from the ADC
  Serial.print("read average: \t\t");
  Serial.println(scale.read_average(20));        // print the average of 20 readings from the ADC
  Serial.print("get value: \t\t");
  Serial.println(scale.get_value(5));            // print the average of 5 readings from the ADC minus the tare weight (not set yet)
  Serial.print("get units: \t\t");
  Serial.println(scale.get_units(5), 1);         // print the average of 5 readings from the ADC minus tare weight (not set) divided

  // by the SCALE parameter (not set yet)
  scale.set_scale(2280.f);                       // this value is obtained by calibrating the scale with known weights; see the README for details
  scale.tare();                                  // reset the scale to 0

  Serial.println("After setting up the scale:");
  Serial.print("read: \t\t");
[...]

PR #116, opened 2022-01-07
Adds HX711 Issue #101

yepher:
This adds the HX711 for Issue #101 

The issue mentions it is for gauge pressure sensor but this is also used for load cells and strain gauges so I made it generic.

### DIFF
--- a/src/hx711-element.stories.ts
+++ b/src/hx711-element.stories.ts
@@ -1,0 +1,17 @@
+import { html } from 'lit';
+import './hx711-element';
+
+export default {
+  title: 'Hx711',
+  component: 'wokwi-hx711',
+  argTypes: {
+    value: { control: { type: 'number', min: -2147483648, max: 2147483647, step: 1 } },
+  },
+  args: {
+    value: 0,
+  },
+};
+
+const Template = () => html`<wokwi-hx711 v</wokwi-hx711>`;
+
+export const Default = Template.bind({});
--- a/src/hx711-element.stories.ts
+++ b/src/hx711-element.stories.ts
@@ -2,16 +2,10 @@ import { html } from 'lit';
 import './hx711-element';
 
 export default {
-  title: 'Hx711',
+  title: 'HX711',
   component: 'wokwi-hx711',
-  argTypes: {
-    value: { control: { type: 'number', min: -2147483648, max: 2147483647, step: 1 } },
-  },
-  args: {
-    value: 0,
-  },
 };
 
-const Template = () => html`<wokwi-hx711 v</wokwi-hx711>`;
+const Template = () => html`<wokwi-hx711></wokwi-hx711>`;
 
 export const Default = Template.bind({});
--- a/src/hx711-element.stories.ts
+++ b/src/hx711-element.stories.ts
@@ -1,19 +1,30 @@
 import { html } from 'lit';
-import { storiesOf } from '@storybook/web-components';
-import { withKnobs } from '@storybook/addon-knobs';
 import './hx711-element';
 
 export default {
   title: 'HX711',
   component: 'wokwi-hx711',
+  argTypes: {
+    angle: { control: { type: '50kg', width: 580, height: 430 } },
+  },
+  args: {
+    angle: 0,
+    units: '',
+    value: '',
+  },
 };
 
-storiesOf('HX711', module)
-  .addParameters({ component: 'wokwi-hx711' })
-  .addDecorator(withKnobs)
-  .add('50kg Load Sensor', () => html`<wokwi-hx711 sensorType="50kg Load Sensor"></wokwi-hx711>`)
-  .add('5kg Load Sensor', () => html`<wokwi-hx711 sensorType="5kg Load Sensor"></wokwi-hx711>`)
-  .add(
-    'Gauge Pressure Sensor',
-    () => html`<wokwi-hx711 sensorType="Gauge Pressure Sensor"></wokwi-hx711>`
-  );
+const Template = ({ type, width, height }) =>
+  html`<wokwi-hx711 .type=${type} .width=${width} .height=${height}></wokwi-hx711>`;
+
+// export const Default = Template.bind({});
+// Default.args = { type: '50kg', width: 580, height: 430 };
+
+export const loadCell50kg = Template.bind({});
+loadCell50kg.args = { type: '50kg', width: 580, height: 430 };
+
+export const loadCell5kg = Template.bind({});
+loadCell5kg.args = { type: '5kg', width: 507, height: 269 };
+
+export const gaugePressure = Template.bind({});
+gaugePressure.args = { type: 'gauge', width: 509, height: 200 };
--- a/src/hx711-element.stories.ts
+++ b/src/hx711-element.stories.ts
@@ -7,18 +7,10 @@ export default {
   argTypes: {
     angle: { control: { type: '50kg', width: 580, height: 430 } },
   },
-  args: {
-    angle: 0,
-    units: '',
-    value: '',
-  },
 };
 
 const Template = ({ type, width, height }) =>
   html`<wokwi-hx711 .type=${type} .width=${width} .height=${height}></wokwi-hx711>`;
-
-// export const Default = Template.bind({});
-// Default.args = { type: '50kg', width: 580, height: 430 };
 
 export const loadCell50kg = Template.bind({});
 loadCell50kg.args = { type: '50kg', width: 580, height: 430 };
--- a/src/hx711-element.stories.ts
+++ b/src/hx711-element.stories.ts
@@ -4,9 +4,6 @@ import './hx711-element';
 export default {
   title: 'HX711',
   component: 'wokwi-hx711',
-  argTypes: {
-    angle: { control: { type: '50kg', width: 580, height: 430 } },
-  },
 };
 
 const Template = ({ type, width, height }) =>
--- a/src/hx711-element.stories.ts
+++ b/src/hx711-element.stories.ts
@@ -6,7 +6,7 @@ export default {
   component: 'wokwi-hx711',
 };
 
-const Template = ({ type, width, height }) =>
+const Template = ({ width, height, type = '50kg' }) =>
   html`<wokwi-hx711 .type=${type} .width=${width} .height=${height}></wokwi-hx711>`;
 
 export const loadCell50kg = Template.bind({});
--- a/src/hx711-element.stories.ts
+++ b/src/hx711-element.stories.ts
@@ -1,4 +1,6 @@
 import { html } from 'lit';
+import { storiesOf } from '@storybook/web-components';
+import { withKnobs } from '@storybook/addon-knobs';
 import './hx711-element';
 
 export default {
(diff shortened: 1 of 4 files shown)


src/hx711-element.stories.ts
```diff
@@ -0,0 +1,17 @@
+import { html } from 'lit';
+import './hx711-element';
+
+export default {
+  title: 'Hx711',
```
urish, line 5: Should be HX711

src/hx711-element.stories.ts
```diff
@@ -0,0 +1,17 @@
...
+  component: 'wokwi-hx711',
+  argTypes: {
+    value: { control: { type: 'number', min: -2147483648, max: 2147483647, step: 1 } },
+  },
+  args: {
+    value: 0,
+  },
+};
+
+const Template = () => html`<wokwi-hx711 v</wokwi-hx711>`;
```
urish, line 15: should be `<wokwi-hx711></wokwi-hx711>` (note the closing tag)

src/hx711-element.stories.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default {
+  title: 'Hx711',
+  component: 'wokwi-hx711',
+  argTypes: {
+    value: { control: { type: 'number', min: -2147483648, max: 2147483647, step: 1 } },
+  },
+  args: {
+    value: 0,
+  },
```
urish, line 12: value is not used, so please remove it

src/hx711-element.ts
```diff
@@ -0,0 +1,491 @@
+import { html, LitElement, svg } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+import { ElementPin, GND, VCC } from './pin';
+
+@customElement('wokwi-hx711')
+export class Hx711Element extends LitElement {
+  @property() value = 0;
```
urish, line 7: this property is not used, so please remove it

src/hx711-element.ts
```diff
@@ -0,0 +1,491 @@
...
+  readonly pinInfo: ElementPin[] = [
+    { name: 'VDD', y: 53, x: 109, number: 1, signals: [VCC()] },
+    { name: 'DOUT', y: 35, x: 109, number: 2, signals: [] },
+    { name: 'SCK', y: 44, x: 109, number: 3, signals: [] },
+    { name: 'GND', y: 26, x: 109, number: 4, signals: [GND()] },
+  ];
+
+  render() {
+    return html`
+      <?xml version="1.0" encoding="UTF-8"?>
```
urish, line 18: this `?xml` declaration may confuse browsers - lets remove it

src/hx711-element.ts
```diff
@@ -0,0 +1,491 @@
...
+      </style>
+      <svg
+        width="34mm"
+        height="20mm"
+        viewBox="0 0 340 200"
+        version="1.1"
+        xmlns="http://www.w3.org/2000/svg"
+        xmlns:xlink="http://www.w3.org/1999/xlink"
+      >
+        <title>HX711</title>
```
urish, line 33: We can remove the title, it's not used anywhere

src/hx711-element.ts
```diff
@@ -0,0 +1,491 @@
...
+            <path d="M143.5,1.5 L203.5,1.5" id="Line"></path>
+            <path d="M205,3 L240,24" id="Line"></path>
+            <path d="M142.5,51.5 L202.5,51.5" id="Line"></path>
+            <path d="M16.5,63.5 L76.5,63.5" id="Line"></path>
+            <path d="M48.5,126.5 L189,126.5" id="Line"></path>
+            <path d="M48.5,137.5 L189,137.5" id="Line"></path>
+            <path d="M16.5,63.5 L0,51" id="Line"></path>
+            <path d="M204,53 L239,74" id="Line"></path>
+            <path d="M143.5,39.5 L203.5,39.5" id="Line"></path>
+            <path d="M205,40 L235,48" id="Line"></path>
```
urish, line 96: The `id` attributes can be remove altogether (as far as I can tell, they are unused, and they way SVG works, if you assign an `id`, it should be unique across the document)

src/hx711-element.ts
```diff
@@ -0,0 +1,491 @@
...
+            <text
+              id="E-"
+              font-family="Arial-BoldItalicMT, Arial"
+              font-size="16"
+              font-style="italic"
+              font-weight="bold"
+              fill="#FBFBFB"
+            >
+              <tspan x="0" y="45">E-</tspan>
+            </text>
```
urish, line 310: All the <text> have the same style, and can be combined into a single `<text>` element with multiple `<tspan>` elements. 

e.g. https://github.com/wokwi/wokwi-elements/blob/c40b1d7482133dc2b418a5e3a05a179b9d7fbc01/src/esp32-devkit-v1-element.ts#L193-L224

src/hx711-element.ts
```diff
@@ -0,0 +1,491 @@
+import { html, LitElement, svg } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+import { ElementPin, GND, VCC } from './pin';
+
+@customElement('wokwi-hx711')
+export class Hx711Element extends LitElement {
```
urish, line 6: Class name should be `HX711Element` (to be consistent with existing elements)

urish (changes requested):
Thanks Chris!

image

What part is this design based on? It looks different than the part in #101, and there's another row of pins not present in the original one. I see you mentioned you made this one generic, how does that work?

Some notes regarding the visual style:
1. We use plain colors for the board itself (it looks like you used a gradient)
2. The holes should be transparent (they should actually be holes in the shape), so that it looks good on either light or dark background
3. We usually avoid drawing the on-board traces (to keep the SVG small). But if we do draw them, the contrast is lower (e.g. [here](https://elements.wokwi.com/?path=/story/hc-sr04--hcsr-04))

Example of an existing part which uses a green board: https://elements.wokwi.com/?path=/story/franzininho--default

Other than that, I left some comments on directly on the code

yepher:
I believe I have all the changes requested in the PR now.

src/hx711-element.ts
```diff
@@ -0,0 +1,185 @@
...
+              <text
+                font-family="Arial-BoldItalicMT, Arial"
+                font-size="16"
+                font-style="italic"
+                font-weight="bold"
+              >
+                <tspan x="0.05859375" y="69" fill="#FBFBFB">A-</tspan>
+                <tspan x="0.4921875" y="18" fill="#FBFBFB">E+</tspan>
+                <tspan x="297.444824" y="42" fill="#FBFBFB">GND</tspan>
+                <tspan x="297.168945" y="66" fill="#FBFBFB">DT</tspan>
```
urish, line 75: Shouldn't it be DOUT?

urish:
Thanks! One more suggestion:

Right now, it looks a little empty, so I thought what we could use the space for. 

How about adding "HX711" as a silk screen print on the board itself?

Perhaps also Load Cell Amp, borrowing from: HX711


Then, we still have the question of the unpopulated pins that connect to the load cells. I think it'll make most sense for now to draw the load cells connected directly to them. WDYT?

src/hx711-element.ts
yepher: I copied the board from this one I have and use. it has `DT` but glad to make the `DOUT`

IMG_0289 2

urish: Looking at google images, it's either DT or DAT. So we can keep DT - but let's make sure the name in pinInfo matches

yepher:
Its empty look is why I originally put some traces :) I will add to the Silkscreen as suggested.

I will add the 50kg Load Cells to the drawing

yepher:
Added the sensors as requested and fixed text and updated silk screen.

urish:
Two more requests:

1. sensorType - let's make the values "5kg", "10kg" and "gauge". less margin for errors (e.g. uppercase vs lowercase). Also, let's make the type of the property more specific than string, e.g. as in the servo element:
https://github.com/wokwi/wokwi-elements/blob/9790133c7ac9a2f77eb66a2beb2aa26ceadd1937/src/servo-element.ts#L16

2. The story is using the old format. It still works, but Storybook may deprecate it at some point. You can find examples of the new format in the Stepper motor element story file (or any element added within the last ~6 months)

yepher:
Added requested changes for resizing view and using newer storyboard format.

src/hx711-element.stories.ts
```diff
@@ -0,0 +1,30 @@
+import { html } from 'lit';
+import './hx711-element';
+
+export default {
+  title: 'HX711',
+  component: 'wokwi-hx711',
+  argTypes: {
+    angle: { control: { type: '50kg', width: 580, height: 430 } },
+  },
```
urish, line 9: can be removed - there's no "angle" for the user to control

src/hx711-element.stories.ts
```diff
@@ -0,0 +1,30 @@
...
+    units: '',
+    value: '',
+  },
+};
+
+const Template = ({ type, width, height }) =>
+  html`<wokwi-hx711 .type=${type} .width=${width} .height=${height}></wokwi-hx711>`;
+
+// export const Default = Template.bind({});
+// Default.args = { type: '50kg', width: 580, height: 430 };
```
urish, line 21: Please remove

src/hx711-element.ts
```diff
@@ -0,0 +1,402 @@
+import { html, LitElement } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+import { ElementPin, GND, VCC } from './pin';
+
+@customElement('wokwi-hx711')
+export class HX711Element extends LitElement {
+  @property() type: '5kg' | '50kg' | 'gauge' | undefined;
+  @property() width = '58mm';
+  @property() height = '43mm';
```
urish, line 9: width/height shouldn't be provided by the user. Instead, they should be derived from the type. 

One way to do this would be to introduce a getter, e.g.:

```ts
get dimensions() {
  switch (this.type) { 
    case '5kg': return {width: '58mm', height:'43mm'};
    ...
  }
}
```

Then, in render, you could do
```ts
 const { width, height } = this.dimensions
```

urish (changes requested):
Thanks for all the changes. We're getting there!

yepher:
Made the requested changes.

src/hx711-element.stories.ts
urish: Did you forget about this one?

urish:
Thank you! 

I did some quick integration test with Wokwi. The only thing that came up - what do we do when the type is invalid? e.g.

image

Right now, the element gets the dimensions of the 50kg sensor, but no sensor is drawn. How about always falling back to the 50kg type if there's no valid type value?

yepher:
Oops, I meant to always fall back to 50kg

src/hx711-element.ts
```diff
@@ -0,0 +1,417 @@
...
+  get dimension() {
+    switch (this.type) {
+      case '50kg':
+        return { width: '580', height: '430' };
+      case '5kg':
+        return { width: '507', height: '269' };
+      case 'gauge':
+        return { width: '509', height: '200' };
+      default:
+        return { width: '580', height: '430' };
```
urish, line 25: why not return width/height as numbers instead of strings? then you can remove the conversion to integer in `render()`

urish:
As far as I can see, the last change only sets the default for storybook, but an invalid value for "type" passed to the element would still result in an empty space.

One way to resolve this in a nice way would be to add a renderSensor() method that will return the relevant svg code based on the value of type, with a "default" fallback case. Something along the lines of:

```ts
renderSensor() {
  switch (this.type) {
     case '5kg':  return svg`<tags here>`;
     case 'guage': return svg`<tags here>`;
     case '50kg':
     default: return svg`<tags here>`;
  }
}
```

Then, in render, you can just replace the whole section with `${this.renderSensor()`

urish:
Hooray!

PR #118, opened 2022-01-14
PC Fan Element - adds to storyboard

yepher:
Adds PC Fan Element. This is based on https://github.com/wokwi/wokwi-features/issues/132


src/index.ts
```diff
@@ -49,3 +49,4 @@ export { MicrosdCardElement } from './microsd-card-element';
 export { DipSwitch8Element } from './dip-switch-8-element';
 export { StepperMotorElement } from './stepper-motor-element';
 export { HX711Element } from './hx711-element';
+export { PcFanElement } from './pc-fan-element';
```
urish, line 52: Should be `PCFanElement`

src/pc-fan-element.stories.ts
```diff
@@ -0,0 +1,40 @@
...
+setInterval(() => {
+  fanAnimation.step();
+  forceReRender();
+}, 10);
+
+export default {
+  title: 'PC Fan',
+  component: 'wokwi-pc-fan',
+  argTypes: {
+    angle: { control: { type: 'number', min: 0, max: 360, step: 1 } },
```
urish, line 27: How about using a "range" type for the control? I think it'll be more user friendly

urish (changes requested):
Looks promising, thank you!

PR #119, opened 2022-01-14
Robot Arm Element - Adds robot arm to storyboard.

yepher:
Adds robot arm

### DIFF
--- a/src/robot-arm-element.stories.ts
+++ b/src/robot-arm-element.stories.ts
@@ -1,0 +1,37 @@
+import { html } from 'lit';
+import './robot-arm-element';
+
+export default {
+  title: 'Robot Arm',
+  component: 'wokwi-robot-arm',
+};
+
+const Template = ({
+  sholderAngle,
+  elbowAngle,
+  wristAngle,
+  gripperAngle,
+  gripperColor,
+  armColor,
+  jointColor,
+}) =>
+  html`<wokwi-robot-arm
+    sholderAngle=${sholderAngle}
+    elbowAngle=${elbowAngle}
+    wristAngle=${wristAngle}
+    gripperAngle=${gripperAngle}
+    gripperColor=${gripperColor}
+    armColor=${armColor}
+    jointColor=${jointColor}
+  ></wokwi-robot-arm>`;
+
+export const Default = Template.bind({});
+Default.args = {
+  sholderAngle: 297,
+  elbowAngle: 52,
+  wristAngle: 56,
+  gripperAngle: 270,
+  gripperColor: '#eeeedd',
+  armColor: '#ffffdd',
+  jointColor: '#111111',
+};
--- a/src/robot-arm-element.ts
+++ b/src/robot-arm-element.ts
@@ -1,0 +1,106 @@
+import { html, LitElement } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+import { ElementPin } from '.';
+
+@customElement('wokwi-robot-arm')
+export class RobotArmElement extends LitElement {
+  /**
+   * Angle of shoulder
+   */
+  @property() sholderAngle = 297;
+
+  /**
+   * Angle of elbow
+   */
+  @property() elbowAngle = 52;
+
+  /**
+   * Angle of wrist
+   */
+  @property() wristAngle = 56;
+
+  /**
+   * Angle of gripper
+   */
+  @property() gripperAngle = 270;
+
+  /**
+   * The color of the gripper
+   */
+  @property() gripperColor = '#eeeedd';
+
+  /**
+   * The color of the arm
+   */
+  @property() armColor = '#ffffdd';
+
+  /**
+   * The color of the joints
+   */
+  @property() jointColor = '#111111';
+
+  readonly pinInfo: ElementPin[] = [
+    { name: 'PWM1', y: 195, x: 25, number: 1, signals: [{ type: 'pwm' }] },
+    { name: 'PWM2', y: 195, x: 30, number: 2, signals: [{ type: 'pwm' }] },
+    { name: 'PWM3', y: 195, x: 35, number: 3, signals: [{ type: 'pwm' }] },
+    { name: 'PWM4', y: 195, x: 40, number: 3, signals: [{ type: 'pwm' }] },
+  ];
+
+  render() {
+    /* eslint-disable prettier/prettier */
+    return html`
+      <style>
+        text {
+          user-select: none;
+        }
+      </style>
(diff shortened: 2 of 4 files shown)


yepher:
I'm not exactly sure how to handle the frame since the robot arm has a fairly wide range of movement.

Is there a way to define the "marching ants" frame independent of the robot arm frame?

src/robot-arm-element.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  readonly pinInfo: ElementPin[] = [
+    { name: 'PWM1', y: 195, x: 25, number: 1, signals: [{ type: 'pwm' }] },
+    { name: 'PWM2', y: 195, x: 30, number: 2, signals: [{ type: 'pwm' }] },
+    { name: 'PWM3', y: 195, x: 35, number: 3, signals: [{ type: 'pwm' }] },
+    { name: 'PWM4', y: 195, x: 40, number: 3, signals: [{ type: 'pwm' }] },
+  ];
+
+  render() {
+    /* eslint-disable prettier/prettier */
```
urish, line 50: why?

src/robot-arm-element.ts
```diff
@@ -0,0 +1,106 @@
...
+   * The color of the arm
+   */
+  @property() armColor = '#ffffdd';
+
+  /**
+   * The color of the joints
+   */
+  @property() jointColor = '#111111';
+
+  readonly pinInfo: ElementPin[] = [
```
urish, line 42: We also need GND/VCC pins (they are not currently used in the simulation, but we do want the schematic to look as close as possible to reality)

src/robot-arm-element.ts
```diff
@@ -0,0 +1,106 @@
...
+              <rect x="45" y="250" width="1" height="9" fill="${this.jointColor}"/>
+              <rect x="52" y="250" width="1" height="9" fill="${this.jointColor}"/>
+          </g>
+          <g transform="translate(46.71 210.57)">
+              <!-- Shoulder -->
+              <g transform="rotate(${this.sholderAngle})">
+                  <path transform="translate(-1.27)" d="m1-36c-19.63 0-36.04 16.39-36.04 36s16.41 36 36.04 36 98.75-9 113.12-10c14.38-1 26.03-11 26.03-26s-11.65-24-26.03-26c-14.37-2-93.49-10-113.12-10z" fill="${this.armColor}" stroke="${this.jointColor}" stroke-miterlimit="10" stroke-width="2"/>
+
+                  <!-- Elbow -->
+                  <g transform="translate(113) rotate(${this.elbowAngle})">
```
urish, line 72: extracting the different values from `this` will probably make the code a bit tidies. i.e. at the start of the function:

```ts
const {elbowAngle, wristAngle, ...} = this;
```

Then you can use the variables directly, without repeating this. every time

src/robot-arm-element.stories.ts
```diff
@@ -0,0 +1,37 @@
+import { html } from 'lit';
+import './robot-arm-element';
+
+export default {
+  title: 'Robot Arm',
+  component: 'wokwi-robot-arm',
```
urish, line 6: How about [wokwi-robotic-arm](https://en.wikipedia.org/wiki/Robotic_arm)? Not being a native English speaker, I'm not sure if there's an actual difference, so I'd love to hear the thoughts of someone whose a native English speaker.

urish (changes requested):
Amazing!

Added some comments / questions.

Also, what is the valid range of angles for each joint? Can we clamp the input angles to make sure we never draw an invalid configuration? (We already have utils/clamp.ts with a simple `clamp()` function)

Regarding the bounding box - is it currently bit enough to contain the entire valid range of movement?

urish:
One more thing - let's use a slider for the angle controls and a color picker for the colors in storybook? 

See the [wokwi-rgb-led story](https://github.com/wokwi/wokwi-elements/tree/master/src/rgb-led-element.stories.ts) for an example for both

src/robot-arm-element.ts
yepher: When I let VS code format the SVG it becomes un-readable and I hate the warnings. But if you really want it to wrap all those lines I can remove it.

src/robot-arm-element.ts
yepher: ok, will add the two pins. I was cheating since I knew they had no function.

src/robot-arm-element.stories.ts
yepher: Works for me

yepher:
I guess I accidently delete this somewhere along the way. I will add it back in:

```

argTypes: {
    sholderAngle: { control: { type: 'number', min: 260, max: 360, step: 1 } },
    elbowAngle: { control: { type: 'number', min: 0, max: 135, step: 1 } },
    wristAngle: { control: { type: 'number', min: 0, max: 120, step: 1 } },
    gripperAngle: { control: { type: 'number', min: 270, max: 360, step: 1 } },
    gripperColor: { control: { type: 'color' } },
    armColor: { control: { type: 'color' } },
    jointColor: { control: { type: 'color' } },
  },
  args: {
    sholderAngle: 270,
    elbowAngle: 82,
    wristAngle: 80,
    gripperAngle: 270,
  },
```

As for yellow, I figured a big heavy dangerous piece of equipment like this robotic arm should be yellow :) Actually I just randomly picked yellow.

urish:
> I guess I accidently delete this somewhere along the way. I will add it back in:

"range" type seems more convenient than "number" for the angles (at least for me)

> As for yellow, I figured a big heavy dangerous piece of equipment like this robotic arm should be yellow :)

haha :) google images show that it's either metallic or black combined with either blue, yellow, light gray, or orange (or sometimes just plain black).

So I guess yellow isn't a bad choice. I thought either metallic or purple could be fun (it's the dominant color in Wokwi), though I've no strong opinion here. Yellow is good too.

src/robot-arm-element.ts
urish: Yes, prettier SVG formatting isn't that great.

But when the choice is between not-so-pleasing but consistent SVG formatting or getting the indentation wrong by the next developer who works on this code, I prefer the former.

Also, prettier should automatically format the code whenever you commit anyway. Didn't it?

src/robot-arm-element.ts
yepher: The more I modify SVG the harder it is to update the SVG if I want to add/remove stuff from it. When it become invalid SVG it is a lot harder to update

yepher:
I was not quite sure how to apply the clamp function since this is an output device.

PR #120, opened 2022-01-30
feat: add relay (KS2E-M-DC5)

Ajtak:
Add relay (KS2E-M-DC5)

### DIFF
--- a/src/ks2e-m-dc5-element.ts
+++ b/src/ks2e-m-dc5-element.ts
@@ -1,0 +1,308 @@
+import { html, LitElement } from 'lit';
+import { customElement } from 'lit/decorators.js';
+import { ElementPin } from './pin';
+
+@customElement('wokwi-ks2e-m-dc5')
+export class KS2EMDC5Element extends LitElement {
+  y1pos = 8.85;
+  y2pos = 70.5;
+  x1pos = 14.7;
+  x2pos = 55;
+  x3pos = 95.3;
+  x4pos = 155.8;
+
+  get pinInfo(): ElementPin[] {
+    return [
+      { name: '1', x: this.x1pos, y: this.y1pos, signals: [], description: '1' },
+      { name: '4', x: this.x2pos, y: this.y1pos, signals: [], description: '4' },
+      { name: '6', x: this.x3pos, y: this.y1pos, signals: [], description: '6' },
+      {
+        name: '8',
+        x: this.x4pos,
+        y: this.y1pos,
+        signals: [{ type: 'power', signal: 'GND' }],
+        description: '8',
+      },
+      { name: '9', x: this.x1pos, y: this.y2pos, signals: [], description: '9' },
+      { name: '11', x: this.x2pos, y: this.y2pos, signals: [], description: '11' },
+      { name: '13', x: this.x3pos, y: this.y2pos, signals: [], description: '13' },
+      { name: '16', x: this.x4pos, y: this.y2pos, signals: [], description: '16' },
+    ];
+  }
+
+  render() {
+    return html`
+      <svg
+        width="44mm"
+        height="21mm"
+        version="1.1"
+        viewBox="0 0 44 21"
+        xmlns="http://www.w3.org/2000/svg"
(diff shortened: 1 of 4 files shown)


src/ks2e-m-dc5-element.ts
```diff
@@ -0,0 +1,308 @@
...
+          >
+            <path
+              transform="scale(.5)"
+              d="m5 0c0 2.76-2.24 5-5 5s-5-2.24-5-5 2.3-5 5-5c2.76 0 5 2.24 5 5z"
+              fill="context-stroke"
+              fill-rule="evenodd"
+              stroke="context-stroke"
+              stroke-width="1pt"
+            />
+          </marker>
```
urish, line 170: What are all those markers? I don't think they are relevant to the drawing, so let's remove them

src/ks2e-m-dc5-element.ts
```diff
@@ -0,0 +1,308 @@
...
+      { name: '13', x: this.x3pos, y: this.y2pos, signals: [], description: '13' },
+      { name: '16', x: this.x4pos, y: this.y2pos, signals: [], description: '16' },
+    ];
+  }
+
+  render() {
+    return html`
+      <svg
+        width="44mm"
+        height="21mm"
```
urish, line 37: Looking at the [data sheet](https://www.datasheetq.com/datasheet-download/752684/1/KEPONT/KS2E-M-DC5), it seems like the size of the element is 21 x 10 x 12mm, so the width should be 21mm and the height should be 10mm. 

This probably means the smaller text ("DC 5V               1A/125V AC               2A/⠀30V DC") will be too small to read - so it's fine to remove it

urish: Also, changing the size means that the positions of the pins in pinInfo will have to be updated

src/ks2e-m-dc5-element.ts
```diff
@@ -0,0 +1,308 @@
...
+import { ElementPin } from './pin';
+
+@customElement('wokwi-ks2e-m-dc5')
+export class KS2EMDC5Element extends LitElement {
+  y1pos = 8.85;
+  y2pos = 70.5;
+  x1pos = 14.7;
+  x2pos = 55;
+  x3pos = 95.3;
+  x4pos = 155.8;
```
urish, line 12: Let's move those outside the class and declare them using const, e.g. 

```js
const y1Pos = 8.85;
const y2Pos = 70.5;
const x1Pos = 14.7;
const x2Pos = 55;
const x3Pos = 95.3;
const x4Pos = 155.8;
```

Then below, just use them without `this.` in front (also note I made the "P" in "Pos" upper case, to keep it consistent with the naming convention of the project)

src/ks2e-m-dc5-element.ts
```diff
@@ -0,0 +1,308 @@
...
+        name: '8',
+        x: this.x4pos,
+        y: this.y1pos,
+        signals: [{ type: 'power', signal: 'GND' }],
+        description: '8',
+      },
+      { name: '9', x: this.x1pos, y: this.y2pos, signals: [], description: '9' },
+      { name: '11', x: this.x2pos, y: this.y2pos, signals: [], description: '11' },
+      { name: '13', x: this.x3pos, y: this.y2pos, signals: [], description: '13' },
+      { name: '16', x: this.x4pos, y: this.y2pos, signals: [], description: '16' },
```
urish, line 29: I think it's better to use descriptive pin names: COIL1, COIL2 for the coil pins, then P1 / NO1 / NC1 (normally open / normally closed), P2, NO2, NC2  for the pole pins.

The pin number can go into the "number" field (e.g. number: 16), and the description can be left out (or just "Common Terminal of POLE1", "Normally Closed Terminal of POLE1"), etc.

urish (changes requested):
Thanks for contributing this element! Please see my comments on the code. Cheers!

Ajtak:
I'm created commit with fixed changes.

urish:
Thank you!

PR #131, opened 2022-07-10
Add 3 digits variant to 7segment-element

DurandA:
As discussed in https://github.com/wokwi/wokwi-features/issues/383, this adds a 3 digits variant to the _7segment_ part.

I was not able to identify a model that was more popular to others, but most Chinese manufacturers seems to use the same pinout ([see LCSC datasheets](https://lcsc.com/products/Led-Segment-Display_532.html)).

Finally, pin 6 is not connected. I am not sure what is the convention for Wokwi in this case.

### DIFF
--- a/src/7segment-element.ts
+++ b/src/7segment-element.ts
@@ -69,6 +69,22 @@ export class SevenSegmentElement extends LitElement {
           { name: 'CLN', ...pinXY(8), signals: [], description: 'Colon' },
         ];
 
+        case 3:
+          // Pinout based on SP420281N model
+          return [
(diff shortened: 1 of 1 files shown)


urish:
Thanks! Your change is now in production :)

> Finally, pin 6 is not connected. I am not sure what is the convention for Wokwi in this case.

Most often, we just add this pin and call it "NC", "NC2", etc. (usually this is how the pins are denoted in the pinout, and we try to keep the names consistent in the pinout). But omitting the pin (like you did) also makes sense to me - no strong opinion which way is better.

PR #136, opened 2022-09-06
Nema Stepper Element

bonnyr:
Added a new element to represent Steppers based on popular Nema sizes

Example images:

image
image


bonnyr:
@AriellaE  any feedback on this PR?

bonnyr:
@AriellaE did you mean you want me to modify the original?

AriellaE:
@bonnyr yes please

bonnyr:
@AriellaE  - modified stepper element as requested (removed the added nema-* variant)

src/stepper-motor-element.ts
```diff
@@ -1,177 +1,126 @@
...
+  }
 
-  readonly pinInfo: ElementPin[] = [
-    { name: 'A+', y: 237, x: 108, number: 1, signals: [] },
-    { name: 'A-', y: 237, x: 98, number: 2, signals: [] },
-    { name: 'B+', y: 237, x: 117, number: 3, signals: [] },
-    { name: 'B-', y: 237, x: 127, number: 4, signals: [] },
-  ];
+  render() { 
+    const spec = this.nemaSpecMap[this.size];
```
AriellaE, line 57: please add fallback behavior when `size` is not valid (e.g. use the values of size 17, or the closest size)

src/stepper-motor-element.ts
```diff
@@ -1,177 +1,126 @@
...
-    return html`
-      <svg
-        width="57mm"
-        height="63.349mm"
-        version="1.1"
-        viewBox="0 0 57 63.349"
-        xmlns="http://www.w3.org/2000/svg"
-        xmlns:xlink="http://www.w3.org/1999/xlink"
+    return html`<svg
+      width="100mm"
```
AriellaE, line 68: SVG size should match the actual size of the graphics. Otherwise, Wokwi will draw the selection rectangle too big when you click on the element.

AriellaE (changes requested):
Thank you. 

It looks good on the functional level.
I reviewed your changes and added some comments.
In addition here are some general requests:
1. please format the code with prettier ( run `npx prettier --write src/**.ts`)
2. match colors to the original stepper so it looks metallic
3. `units` and `value` texts should go in separate lines, just like the original stepper. the value should have a bigger font.

Bonus: try to make the center of rotation more accurate, so the rotator won't shift as you rotate it.

src/stepper-motor-element.ts
```diff
@@ -1,177 +1,126 @@
...
-    { name: 'A-', y: 237, x: 98, number: 2, signals: [] },
-    { name: 'B+', y: 237, x: 117, number: 3, signals: [] },
-    { name: 'B-', y: 237, x: 127, number: 4, signals: [] },
-  ];
+  render() { 
+    const spec = this.nemaSpecMap[this.size];
 
-  render() {
-    const { arrow } = this;
+    // local vars    const ratio = 1; //100 / spec.frameSize;
```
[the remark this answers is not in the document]
bonnyr: done in latest commit

src/stepper-motor-element.ts
bonnyr: isn't this what happens when specifying the property as one of a set of values?

src/stepper-motor-element.ts
bonnyr: are you suggesting that I should vary the generated svg to match the size of the component?
I can do that

AriellaE (changes requested):
It looks very good, thank you!

Few more comments: 
1. Text size of `value` must always be larger than `units` (the proportions should be around 14:24) , and now that there are size option to the stepper itself, the texts should grow and shrink as the stepper.  
2. In storybook please cancel the rotation animation(although it very pretty) . You can create a new story that contains the animation, but all the other stories (including default  story) should be static with the ability of controlling the angle.

bonnyr:
> It looks very good, thank you!
great :)

> 
> Few more comments:
> 
> 1. Text size of `value` must always be larger than `units` (the proportions should be around 14:24) , and now that there are size option to the stepper itself, the texts should grow and shrink as the stepper.

Modified the `units` property text size to be 2/3 of the `value` .  The text size itself is already a function of the selected instance size (see `textSize`  property value in the NEMASpec class)

> 2. In storybook please cancel the rotation animation(although it very pretty) . You can create a new story that contains the animation, but all the other stories (including default  story) should be static with the ability of controlling the angle.

Removed animation altogether. For some reason it does not want to work when I apply to Template type stories (all other stories using animation make direct use of `storiesOf` and friends)

Hope this is it...

bonnyr:
@AriellaE  - any update?

AriellaE:
Hi @bonnyr  
I merged the changes released new version and uploaded it to wokwi.com.
Thanks again for contributing to Wokwi!
Shana tova 🍯🍎

PR #137, opened 2022-09-08
feature: VID2805 Dual Stepper Element

bonnyr:
Adding a new element to represent a common dual stepper - the VID28-05

The part looks like this:
image

The component looks like this (top view):
image

and with shorter hands:
image

The part has two sets of the typical stepper pins (A-/A+/B-/B+)

One note is that I am not sure whether the pinInfo coords match the position in the drawing or how to make sure they do.
Another note is that on storybook, I am not sure how to create two group of controls, one for each stepper hand and be able to use these controls when initializing a class representing each of the hands


urish:
Error message from Netlify:

```
4:11:04 AM: src/vid2805-dual-stepper-element.ts(88,27): error TS2550: Property 'replaceAll' does not exist on type 'string'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2021' or later.
4:11:04 AM: src/vid2805-dual-stepper-element.ts(89,27): error TS2550: Property 'replaceAll' does not exist on type 'string'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2021' or later.
```

You can probably see the same errors if you run `npm run build` locally.

`replaceAll` is newer JavaScript function, and isn't supported yet by all browsers. An alternative would be to use `someString.split("search").join("replace")`

bonnyr:
Thanks for the clarification @urish 

This now builds correctly.
There's the issues of:

1. I am not sure whether the pinInfo coords match the position in the drawing or *how to make sure they do*.
2. In storybook, I am not sure how to create two group of controls, one for each stepper hand and be able to use these controls when initializing a class representing each of the hands - currently the controls do not really translate into a change in the selected template

Any help here would be appreciated

urish:
> I am not sure whether the pinInfo coords match the position in the drawing or how to make sure they do.

Look at the [Debugging Pin Info section of the Contributing guide](https://github.com/wokwi/wokwi-elements/blob/master/CONTRIBUTING.md#debugging-pin-info) to learn how to test it

> In storybook, I am not sure how to create two group of controls, one for each stepper hand and be able to use these controls when initializing a class representing each of the hands - currently the controls do not really translate into a change in the selected template

I'm not sure if storybook supports groups of controls - you can take a quick look at their docs / examples and see if they do. Otherwise, you can probably use a prefix to separate the controls (e.g. `hand1Something` and `hand2Something`). Once I have an opportunity to look at the your code (probably some time next week), I will have more context to help.

bonnyr:
@urish pinInfo stumps me...

The SVG for the element is using specific size in mm, the pins are drawn using relative path operations and are scaled by `mmToPix` 

When returning pin info with the x coordinate the same as that specified in the path, the debug view shows the dots too close together:
```
  get pinInfo(): ElementPin[] {
    const { x, y, innerOff, outerOff, trY} = this.coords();

    return [
    { name: 'A1+', y: trY, x: 58 * 1, number: 1, signals: [] },
    { name: 'A1-', y: trY, x: 65  * 1, number: 2, signals: [] },
    { name: 'B1+', y: trY, x: 72  * 1, number: 3, signals: [] },
```
image

When scaling the information in pinInfo by `mmToPix` the dots are too spaced too wide apart:
```
 get pinInfo(): ElementPin[] {
    const { x, y, innerOff, outerOff, trY} = this.coords();

    return [
    { name: 'A1+', y: trY, x: 58 * mmToPix, number: 1, signals: [] },
    { name: 'A1-', y: trY, x: 65  * mmToPix, number: 2, signals: [] },
  ```
image


I've pushed the code that depicts that - any help would be appreciated.

**Comment by bonnyr** [comment]
@urish - managed to get the component and pins aligned and correct size - are you able to look at the component and accept/reject?

File: src/vid2805-dual-stepper-element.ts [hunk 1]
```diff
@@ -0,0 +1,191 @@
...
+    inner_plain_hand:
+      'm0,0 c 0 2.6 2.1 4.7 4.7 4.7 c 1.27 0 2.41 -0.5 3.25 -1.31 h 0 c 0.5 -0.49 1.1 -0.89 1.8 -1.19 c 0.7 -0.3 1.4 -0.4 2.1 -0.4 h ${len} c 1 0 1.8 -0.8 1.8 -1.8 s -0.8 -1.8 -1.8 -1.8 h -${len} s -0.06 0 -0.06 0 c -0.7 0 -1.5 -0.1 -2.1 -0.4 c -0.7 -0.3 -1.3 -0.7 -1.8 -1.2 h 0 c -0.84 -0.8 -1.98 -1.3 -3.24 -1.3 c -2.6 0 -4.7 2.1 -4.7 4.7 Z m 2.7 0 c 0 -1.1 0.9 -2 2 -2 s 2 0.9 2 2 s -0.9 2 -2 2 s -2 -0.9 -2 -2 Z',
+    inner_arrow_hand:
+      'm 0 0 c 0 2.6 2.1 4.7 4.7 4.7 c 1.37 0 2.59 -0.59 3.44 -1.52 c 0.7 -0.7 1.5 -1.2 2.5 -1.5 s 2 -0.3 2.9 -0.1 l 0.9 0.2 l ${len} -1.8 l -${len} -1.8 l -0.9 0.2 c -1 0.2 -2 0.2 -2.9 -0.1 c -1 -0.3 -1.8 -0.8 -2.5 -1.5 c -0.86 -0.93 -2.08 -1.52 -3.44 -1.52 c -2.6 0 -4.7 2.1 -4.7 4.7 z m 2.7 0 c 0 -1.1 0.9 -2 2 -2 s 2 0.9 2 2 s -0.9 2 -2 2 s -2 -0.9 -2 -2 z',
+  };
+
+  render() {
+    let inner_svg = this.handMap['inner_' + this.innerHand.shape + '_hand'];
+    let outer_svg = this.handMap['outer_' + this.outerHand.shape + '_hand'];
```
urish, line 96: should be `innerSvg` and `outerSvg` (camelCase)

src/vid2805-dual-stepper-element.ts
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    inner_svg = inner_svg.split('${len}').join(this.innerHand.length.toString());
+    outer_svg = outer_svg.split('${len}').join(this.outerHand.length.toString());
+
+    const { x, y, innerOff, outerOff, trY } = this.coords();
+    console.debug(this.coords());
+    console.debug(trY);
+    console.debug(x);
+    console.debug(y);
```
urish, line 115: Please remove these

src/vid2805-dual-stepper-element.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    return html`
+    <svg id="VID2805" xmlns="http://www.w3.org/2000/svg" width=70mm height=70mm viewBox="0 0 265 265">
```
urish, line 118: Please remove the `id` from the `<svg>`

src/vid2805-dual-stepper-element.ts
```diff
@@ -0,0 +1,191 @@
...
+          <use href="#pin-1" x=2.54 />
+          <use href="#pin-1" x=5.08 />
+          <use href="#pin-1" x=7.62 />
+          <use href="#pin-1" x=10.16 />
+          <use href="#pin-1" x=12.70 />
+          <use href="#pin-1" x=15.24 />
+          <use href="#pin-1" x=17.78 />
+        </g>
+
+        <g id="scaled_body" transform="scale(1.333)">
```
urish, line 150: ids should use kebab-case, i.e. `scaled-body`. In general, if you don't use an id, you can remove it

src/vid2805-dual-stepper-element.ts
```diff
@@ -0,0 +1,191 @@
...
+export class StepperHand {
+  /**
+   * length used to control hand length
+   */
+  length = 25;
+
+  /**
+   * colour used to paint the hand
+   */
+  colour = '';
```
urish, line 17: Let's use `color` to be consistent with the other elements (e.g. the LED, pushbutton, etc.)

src/vid2805-dual-stepper-element.ts
```diff
@@ -0,0 +1,191 @@
...
+import { ElementPin } from '.';
+import { mmToPix } from './utils/units';
+
+export type HandShape = 'arrow' | 'plain'; // | 'fancy';
+
+export class StepperHand {
+  /**
+   * length used to control hand length
+   */
+  length = 25;
```
urish, line 12: What are the units here?  please add to the comment.

I think millimeters makes most sense from user perspective.

src/vid2805-dual-stepper-element.ts
```diff
@@ -0,0 +1,191 @@
+import { html, LitElement, svg } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+import { ElementPin } from '.';
+import { mmToPix } from './utils/units';
+
+export type HandShape = 'arrow' | 'plain'; // | 'fancy';
```
urish, line 6: Remove 'fancy'

src/vid2805-dual-stepper-element.ts
```diff
@@ -0,0 +1,191 @@
...
+      { name: 'A1-', ...pinXY(1), number: 2, signals: [] },
+      { name: 'B1+', ...pinXY(2), number: 3, signals: [] },
+      { name: 'B1-', ...pinXY(3), number: 4, signals: [] },
+      { name: 'A2+', ...pinXY(4), number: 5, signals: [] },
+      { name: 'A2-', ...pinXY(5), number: 6, signals: [] },
+      { name: 'B2+', ...pinXY(6), number: 7, signals: [] },
+      { name: 'B2-', ...pinXY(7), number: 8, signals: [] },
+    ];
+
+    console.debug(pi);
```
urish, line 79: remove

src/vid2805-dual-stepper-element.ts
```diff
@@ -0,0 +1,191 @@
...
+      'm0,0 c 0 2.6 2.1 4.7 4.7 4.7 c 1.27 0 2.41 -0.5 3.25 -1.31 h 0 c 0.5 -0.49 1.1 -0.89 1.8 -1.19 c 0.7 -0.3 1.4 -0.4 2.1 -0.4 h ${len} c 1 0 1.8 -0.8 1.8 -1.8 s -0.8 -1.8 -1.8 -1.8 h -${len} s -0.06 0 -0.06 0 c -0.7 0 -1.5 -0.1 -2.1 -0.4 c -0.7 -0.3 -1.3 -0.7 -1.8 -1.2 h 0 c -0.84 -0.8 -1.98 -1.3 -3.24 -1.3 c -2.6 0 -4.7 2.1 -4.7 4.7 Z m 2.7 0 c 0 -1.1 0.9 -2 2 -2 s 2 0.9 2 2 s -0.9 2 -2 2 s -2 -0.9 -2 -2 Z',
+    inner_arrow_hand:
+      'm 0 0 c 0 2.6 2.1 4.7 4.7 4.7 c 1.37 0 2.59 -0.59 3.44 -1.52 c 0.7 -0.7 1.5 -1.2 2.5 -1.5 s 2 -0.3 2.9 -0.1 l 0.9 0.2 l ${len} -1.8 l -${len} -1.8 l -0.9 0.2 c -1 0.2 -2 0.2 -2.9 -0.1 c -1 -0.3 -1.8 -0.8 -2.5 -1.5 c -0.86 -0.93 -2.08 -1.52 -3.44 -1.52 c -2.6 0 -4.7 2.1 -4.7 4.7 z m 2.7 0 c 0 -1.1 0.9 -2 2 -2 s 2 0.9 2 2 s -0.9 2 -2 2 s -2 -0.9 -2 -2 z',
+  };
+
+  render() {
+    let inner_svg = this.handMap['inner_' + this.innerHand.shape + '_hand'];
+    let outer_svg = this.handMap['outer_' + this.outerHand.shape + '_hand'];
+
+    // todo: validate hands
```
urish, line 98: Please do!

The expected behavior is that the element should still render if the inputs are invalid (i.e. use some sensible default value)

src/vid2805-dual-stepper-element.ts
```diff
@@ -0,0 +1,191 @@
...
+      { name: 'B2+', ...pinXY(6), number: 7, signals: [] },
+      { name: 'B2-', ...pinXY(7), number: 8, signals: [] },
+    ];
+
+    console.debug(pi);
+    return pi;
+  }
+
+  readonly handMap: { [key: string]: string } = {
+    outer_plain_hand:
```
urish, line 84: key names should be camelCase, e.g. `outerPlainHand`.

src/vid2805-dual-stepper-element.ts
```diff
@@ -0,0 +1,191 @@
...
+      'm0,0 c0,5,4,9,9,9,3.28,0,6.13-1.73,7.7-4.33v.03c.5-.8,1.2-1.6,2.1-2.1,.8-.5,1.8-.8,2.8-.8h${len}c1,0,1.8-.8,1.8-1.8s-.8-1.8-1.8-1.8h-${len}c-1,0-1.9-.3-2.8-.8-.8-.5-1.6-1.2-2.1-2h0c-1.56-2.64-4.43-4.4-7.74-4.4-5,0-9,4.1-9,9Zm3.5,0c0-3,2.4-5.4,5.5-5.4s5.5,2.4,5.5,5.4-2.5,5.4-5.5,5.4-5.5-2.4-5.5-5.4Z',
```
urish, line 85: The use of `${len}` can be confusing here, as the same syntax is used for JavaScript's template strings. 

I suggest to turn these into a function that actually uses JavaScript template strings for the substitution, e.g.: 

```js
outerPlainHand: (len: number) => 
   `m0,0 c0,5,4,9,9,9,3.28,0,6.13-1.73,7.7-4.33v.03c.5-.8,1.2-1.6,2.1-2.1,.8-.5,1.8-.8,2.8-.8h${len}c1,0,1.8-.8,1.8-1.8s-.8-1.8-1.8-1.8h-${len}c-1,0-1.9-.3-2.8-.8-.8-.5-1.6-1.2-2.1-2h0c-1.56-2.64-4.43-4.4-7.74-4.4-5,0-9,4.1-9,9Zm3.5,0c0-3,2.4-5.4,5.5-5.4s5.5,2.4,5.5,5.4-2.5,5.4-5.5,5.4-5.5-2.4-5.5-5.4Z`,
```

src/vid2805-dual-stepper-element.stories.ts
```diff
@@ -0,0 +1,88 @@
...
+  component: 'wokwi-vid2805-dual-stepper',
+  argTypes: {
+    outerLength: { control: { type: 'range', min: 20, max: 70 } },
+    outerAngle: { control: { type: 'range', min: 0, max: 360 } },
+    outerColour: { control: { type: 'color' } },
+    outerShape: { options: ['arrow', 'plain', 'fancy'], control: { type: 'select' } },
+    innerLength: { control: { type: 'range', min: 20, max: 70 } },
+    innerAngle: { control: { type: 'range', min: 0, max: 360 } },
+    innerColour: { control: { type: 'color' } },
+    innerShape: { options: ['arrow', 'plain', 'fancy'], control: { type: 'select' } },
```
urish, line 16: Seems like all these controls are non functional - see my comment on the `outerHand` / `innerHand` properties in `vid2805-dual-stepper-element.ts` below.

Also, `fancy` is not implements for the shapes, so please remove it

src/vid2805-dual-stepper-element.ts
```diff
@@ -0,0 +1,191 @@
...
+    this.angle = angle;
+    this.colour = colour;
+    this.shape = shape;
+    this.length = length;
+  }
+}
+
+@customElement('wokwi-vid2805-dual-stepper')
+export class VID2805DualStepperElement extends LitElement {
+  @property() outerHand = new StepperHand();
```
urish, line 39: Encapsulating the hand properties in an object is neater, but I'm afraid it causes two issues:

1. It makes the element unusable from HTML. You always have to use JavaScript to create the `innerHand`/`outerHand` objects.
2. lit is unable to track the properties inside `innerHand`/`outerHand`. Thus, if you change any of the properties (e.g. `document.querySelector('wokwi-vid2805-dual-stepper').innerHand.angle = 40`), it won't pick up the change. The only way is to create a new `innerHand` object every time you want to change a property.

So I think that in this situation and given the limitations of the technology (lit + html), we should change the design to have individual properties per each hand attribute (`innerAngle`, `innerColor`, etc). This will also fix the issue with the story controls not working.

src/vid2805-dual-stepper-element.ts
```diff
@@ -0,0 +1,191 @@
...
+            <stop stop-color="#4f4f4f" offset="-25%" />
+            <stop stop-color="grey" offset="75%" />
+        </linearGradient>
+    </defs>
+      <g id="vid2805-translated" transform="translate(15,265) rotate(270 0 0) translate(0,${trY})">
+
+      <!-- pins - translate to top of base, between holes -->
+        <g id="pins" transform="scale(${mmToPix}) translate(21 3) ">
+          <path id="pin-1" class="cls-3" d="m 0 0 v -3 c 0 -1 1 -1 1 0 v 3 z" />
+          <use href="#pin-1" x=2.54 />
```
urish, line 141: Need to surround attribute values with quotes ("), otherwise, some browsers (Safari, if I recall correctly) are unhappy and won't render the SVG

urish (changes requested):
Thanks for working on this! 

I left a bunch of comments. The most important point is regarding the `innerHand` / `outerHand` attribute. The other points are mostly regarding naming convention, removing debug prints, improving code readability, etc.

One more important point is the name of the element. I looked at the [datasheet](http://vidmotion.com.cn/UploadFile/down/2009111390206_VID28%20manual%20EN-080623.pdf), and it seems like the real part is called `vid28-05` and not `vid2805`, even though the datasheet describes it as `vid28-xx`. 

I suggest to go with `wokwi-vid28-dual-stepper`. When you google "VID28" it comes up with images of this specific stepper motor, so the term "vid28" seems to be specific enough (and I prefer shorter names where possible).

src/vid2805-dual-stepper-element.ts
bonnyr: This is great - I was not aware I could do this!

src/vid2805-dual-stepper-element.ts
bonnyr: renamed as requested

src/vid2805-dual-stepper-element.ts
bonnyr: done

src/vid2805-dual-stepper-element.ts
bonnyr: done

src/vid2805-dual-stepper-element.ts
bonnyr: done

src/vid2805-dual-stepper-element.ts
bonnyr: done

src/vid2805-dual-stepper-element.stories.ts
bonnyr: done - removed composite controls and now using only individual properties

src/vid2805-dual-stepper-element.ts
bonnyr: as discussed, removed

src/vid2805-dual-stepper-element.ts
bonnyr: done

src/vid2805-dual-stepper-element.ts
bonnyr: commend indicates now the length is in mm

src/vid2805-dual-stepper-element.ts
bonnyr: this was a placeholder for an ornate hand paths that I was in the process of adding. Now added

urish:
Thanks! "ornate" is cool. I'll hope to go over the actual changes soon!

urish:
Thanks again for making the changes!

Looking at it, here are some issues which will affect the usability on Wokwi:
1. The bounding SVG element is too big. We need it to be as tight as possible, as Wokwi uses this when drawing the selection rectangle around the element.
image

2. The element and pins moves when you change the arm length. Let's horizontally center the element within the SVG, so it stays in the same places (and the pins too), regardless of the hand length.

If we made the maximum hand length shorter (e.g. 50 instead of 70), we'd need to leave less padding around the element. I'm not super concerned about the current amount of padding, so if there's a good reason to keep it at 70, that's fine too.

bonnyr:
> The bounding SVG element is too big. We need it to be as tight as possible, as Wokwi uses this when drawing the selection rectangle around the element.

and 
> The element and pins moves when you change the arm length. Let's horizontally center the element within the SVG, so it stays in the same places (and the pins too), regardless of the hand length.

I had to make this as large as it is to account for the longer arms - they extend beyond the base.
The use case for making this element this large is when placing a number of these vertically - the hands then forms circles that are closer together.

If this is not compatible with the way Wokwi works, I am happy with your suggestion of reducing the maximum size.
This will also solve the movement issue since the arms will be confined to the bounding box of the element itself and we can make the element as tight as possible.

Thoughts?

urish:
> I had to make this as large as it is to account for the longer arms - they extend beyond the base.

Right now, there's extra margin on the right side - the photo that I took in the previous has the largest arm size.

I see two ways to go about this:
1. Reducing the SVG size so it has just enough padding to fit tightly when the arms are in their largest size (perhaps reducing the max arm length a bit, so we need less padding)
2. Introducing some mechanism to tell Wokwi about the true bounding box of the element (similar to what we have in `pinInfo`). That will probably also be useful for #119 if it ever lands.

The first option will be faster, as the second one needs some changes on the Wokwi side as well.

In any case, the pin positions / motor body should stay static even when you change the arm length.

I hope this all makes sense to you.

bonnyr:
@urish - another update, addressing the comments I believe

PR #140, opened 2022-10-02
updated based on working with Wokwi

bonnyr:
changed A pins polarity to match 4988 wiring layout, widened the plain/arrow hands so they look more substantial


urish:
Thanks! Change is now up on wokwi.com

PR #143, opened 2022-11-18
feat: add neopixel element

minolmal:
- added GLCD12864 element (followed SSD1306 element)

known errors: canvas size and display size does not match. I need a little help here to figure this out.


urish:
Thanks! The graphics looks nice.

For screens, we're now adding them through [wokwi-boards](https://github.com/wokwi/wokwi-boards). Here's an example for the SSD1306: https://github.com/wokwi/wokwi-boards/tree/master/boards/ssd1306

However, in addition to the graphics, we'd need to implement the controller for the chip. Here's an example for a custom chip implementation of a display controller: https://wokwi.com/projects/333332561949360723

Issue #149, opened 2023-01-31
Add .js extension to export/import of javascript files in npm package


jogibear9988:
Hello,

I'm woring on a designer (https://node-projects.github.io/web-component-designer-demo/index.html) wich can directly import webcomponents from npm modules wich use esm code. But your npm packges is missing .js extension in the package, could that be added?

urish:
Thanks for the request. Unfortunately, we don't have the bandwidth to look into this.

PR #150, opened 2023-02-03
fix issue #149 - add .js to esm modules

jogibear9988:


jogibear9988:
so the elements could be used in https://node-projects.github.io/web-component-designer-demo/index.html

urish:
Thanks. Could the @node-projects/add-js-to-esm-modules be modified to produce less output? right now it's very chatty. Ideally it'd produce no output, or just a summary with the number of files/import statements modified and the run time.

Also, is it possible to add some unit tests to the source code of that module? To ensure that future version won't cause regressions that will possibly break the build of packages using it.

jogibear9988:
will do some tests and be less chatty 😉
but will do it next week.

urish:
@GervinFung congrats on your projects! This is not really a priority for me now, but I'll keep it in mind.

